PR #64, opened 2021-10-28
Better printing for unbound args

tkf:


lgoettgens:
Similar changes have already been incorporated in https://github.com/JuliaTesting/Aqua.jl/pull/146.

PR #71, opened 2022-02-24
implement ambiguity hint

Roger-luo:
resolve #52

PR #73, opened 2022-03-25
Fix `UndefVarError` in Julia v1.0-1.3

giordano:
Ref: https://github.com/JuliaTesting/Aqua.jl/pull/71/files#r834842406.  If you have better ideas about how to more proper fix this or how to actually test this code path which is clearly untested, please take over this PR or open a new one, but at least this doesn't break the package in older versions of Julia.  Not very useful to have a QUality Assurance package that breaks your tests.

CC:  @Roger-luo

### DIFF
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -195,8 +195,10 @@ function test_ambiguities_impl(
         println("Ambiguity #", i)
         println(m1)
         println(m2)
-        ambiguity_hint(m1, m2)
-        println()
+        @static if isdefined(Base, :morespecific)
+            ambiguity_hint(m1, m2)
+            println()
+        end
         println()
     end
     return ambiguities == []


tkf:
Thanks! It'd be nice to actually hit this case in the test but let's register this version first.

If you have some simple-ish examples for exercising #71, it'd be great if you can add some in https://github.com/JuliaTesting/Aqua.jl/blob/master/test/pkgs/PkgWithAmbiguities.jl (cc @Roger-luo)

Roger-luo:
Yeah I'm not sure how to test the checks tho, but I can have a look at that morespecific function

tkf:
That'd be great. But I'm also OK with just doing a "smoke test" to hit the code using `morespecific` with some MWE functions. It won't be a comprehensive test but it's still useful since `morespecific` is an internal function. (And many of Aqua's tests are smoke tests anyway.)

giordano:
For what is worth, I found this error in https://github.com/JuliaPhysics/Measurements.jl/pull/82, but I have little clue of where the ambiguity comes from: tests are passing on `master`, and that PR is only changing the signature of an internal macro, in addition to loading a new package (`FiniteDifferences` instead of `Calculus`).  Can the ambiguity come from the new package?

Thanks for the quick merge!  And sorry for the salty comment above, but a failure from a testing package triggered me yesterday night :slightly_smiling_face:

tkf:
Oh, no, that's OK :) I totally missed that my test didn't hit the code.

Issue #83, opened 2022-08-03
Aqua.test_all triggers all deprecation warnings


visr:
Recently GeometryBasics.jl added `Aqua.test_all` in `runtests.jl`. It is working well, but the package has quite a few deprecations defined in [deprecated.jl](https://github.com/JuliaGeometry/GeometryBasics.jl/blob/3cdb91f518db3196b2dedea2a097233ca4778eb3/src/deprecated.jl).

Since [adding Aqua](https://github.com/JuliaGeometry/GeometryBasics.jl/commit/3cdb91f518db3196b2dedea2a097233ca4778eb3) I noticed the full list of deprecation warnings showing up each time the tests are run, here are the last few of them:

```
WARNING: GeometryBasics.Vec3f0 is deprecated, use Vec3f instead.
  likely near D:\visser_mn\.julia\dev\GeometryBasics\test\runtests.jl:7
WARNING: GeometryBasics.Vec4f0 is deprecated, use Vec4f instead.
  likely near D:\visser_mn\.julia\dev\GeometryBasics\test\runtests.jl:7
WARNING: GeometryBasics.Vecf0 is deprecated, use Vecf instead.
  likely near D:\visser_mn\.julia\dev\GeometryBasics\test\runtests.jl:7
Test Summary:  | Pass  Broken  Total   Time
GeometryBasics |  339       2    341  30.1s
     Testing GeometryBasics tests passed
```

Uncommenting `Aqua.test_all` makes these warnings disappear. It's not a serious issue, but it would be nice if these would not show up, since it gives the impression that deprecated bindings are being used during the tests.

PR #89, opened 2022-11-15
do not trigger deprecation warnings

t-bltg:
`getproperty` triggers deprecation warnings: avoid triggering those when walking modules.

Fix https://github.com/JuliaTesting/Aqua.jl/issues/83.

Also observed in the `UnicodePlots` [test suite](https://github.com/JuliaPlots/UnicodePlots.jl/actions/runs/3465118212/jobs/5787474874#step:6:337), unfortunately this PR only fixes `GeometryBasics` tests mentioned in the issue, we also need a PR in `Reexports.jl` or `Colors.jl`.

### DIFF
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -1,8 +1,9 @@
 function walkmodules(f, x::Module)
     f(x)
-    for n in names(x; all = true)
-        # `getproperty` triggers deprecation warnings
-        if !Base.isdeprecated(x, n) && isdefined(x, n)
+    for n in names(x; all = true, imported = true)
+        # `isdefined` and `getproperty` can trigger deprecation warnings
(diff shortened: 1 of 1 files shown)


src/exports.jl
```diff
@@ -1,7 +1,9 @@
 function walkmodules(f, x::Module)
     f(x)
-    for n in names(x; all=true)
-        if isdefined(x, n)
+    for n in names(x; all = true, imported = true)
```
tkf, line 3: Since we check `parentmodule(y) === x` below, `imported = true` here does not seem to be required (and causes extra works).  Or is there a reason that I miss for adding this?

```suggestion
    for n in names(x; all=true)
```

tkf: I'll go ahead and merge this for now. Thanks for the patch!

t-bltg: Thanks !

Issue #94, opened 2023-01-02
broken link in test_piracy


dpinol:
`[test_piracy(testtarget)](https://juliatesting.github.io/Aqua.jl/stable/@ref%20test_piracy)` link at https://juliatesting.github.io/Aqua.jl/stable/ is broken.

I did not manage to fix it specifying the module at Aqua.jl Maybe https://github.com/JuliaDocs/Documenter.jl/pull/1352 fixes it?

It would be nice also to expand a bit the documentation text, or maybe add a link https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy

PR #97, opened 2023-01-25
Add `Aqua.Piracy` to autodocs block

hyrodium:
This PR fixes https://github.com/JuliaTesting/Aqua.jl/issues/94.


hyrodium:
Thank you too for maintaining this package!

PR #101, opened 2023-03-28
Check for piracy against Base by default

jakobnissen:
Previously, libraries which pirated Base, but did not manually bring the offending function into scope was not detected.
Now, default to checking Base for being a victim of piracy by default.

Edit: CI "fails" due to codecoverage acting up.

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -35,13 +35,16 @@ function all_methods!(
     result
 end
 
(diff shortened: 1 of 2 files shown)


lgoettgens:
This feature has been included in https://github.com/JuliaTesting/Aqua.jl/pull/156

PR #102, opened 2023-03-28
Always use Base implementation of nameof(::DataType)

### LINKED ISSUE #95: Something is broken in piracy check
We have recently started to get the following error in our CI (previously everything was OK): 

https://github.com/biaslab/ReactiveMP.jl/actions/runs/3929126548/jobs/6717570554

This error suggests that some code in Aqua.jl does not work well for the following definition:
```julia
Base.nameof(::Type{T}) where {N, R, V <: NTuple{N, <:R}, T <: ManyOf{V}} = string("ManyOf{", N, ", ", nameof(dropproxytype(R)), "}")
```

There is no type-piracy, because the type `T` explicitly set to be of type `ManyOf`, which is defined in our package. The error says that `N` is not defined, but it is defined in the code and this function also executes fine without any errors. That makes me think that type-piracy logic in Aqua is incomplete, but I do not understand the implementation and I'm not quite sure where the issue is.

jakobnissen:
Fixes #95

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -84,7 +84,7 @@ is_foreign_module(mod::Module, pkg::Base.PkgId) = Base.PkgId(mod) != pkg
 function is_foreign(@nospecialize(T::DataType), pkg::Base.PkgId)
     params = T.parameters
     # For Type{Foo}, we consider it to originate from the same as Foo
-    C = getfield(parentmodule(T), nameof(T))
+    C = getfield(parentmodule(T), invoke(nameof, Tuple{DataType}, T))
     if C === Type
         @assert length(params) == 1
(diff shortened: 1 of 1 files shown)


fingolfin:
Is this really necessary, given that #95 seems to have been resolved another way?

If it is, where does it stop, couldn't any number of other methods potentially be overridden and then "require" such a treatment? That doesn't seem sustainable, so perhaps better to just say "if you do this kind of stuff, all bets are off"

Maybe I am missing something that makes `nameof` special in this regard, though?

fingolfin:
Thank you for your contribution! But I think at this time we should not apply this, at least not without a deeper understanding why `nameof` is special compared to a bunch of other functions for which packages might install "bad" methods.

PR #113, opened 2023-05-22
Make analyze_stale_deps more robust

### LINKED ISSUE #60: Printing at `__init__` time throws off stale dependencies check
Specifically, https://github.com/JuliaTesting/Aqua.jl/blob/9eff6edf048720f0d32900bc30730570e31e305e/src/stale_deps.jl#L57

throws. Generally printing during `__init__` isn't good practice since downstream packages will print when they load, but it can be helpful for leaf packages (i.e. applications).

fingolfin:
Otherwise, if any of the loaded packages prints during loading, this messes up things.

Real-world example: GAP.jl, where there is a `\r` being printed by some C code it loads; we'll try to fix that there, but it seems prudent to add this bit of defensive coding regardless of that.

Resolves #60

### DIFF
--- a/src/stale_deps.jl
+++ b/src/stale_deps.jl
@@ -44,9 +44,11 @@ function _analyze_stale_deps_1(pkg::PkgId; ignore::AbstractArray{Symbol} = Symbo
     end
     deps = [PkgId(UUID(v), k) for (k, v) in raw_deps]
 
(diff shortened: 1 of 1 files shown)


lgoettgens:
I think this closes https://github.com/JuliaTesting/Aqua.jl/issues/60.

PR #115, opened 2023-05-23
Check `docs/Project.toml` as well, if it exists

### LINKED ISSUE #51: Test docs project for missing compat?
It would be great if Aqua could test other Project.toml's for missing compat entries such as the one in the docs folder. That is a slightly tricky case since PkgX/docs likely has PkgX as a dep but does not want a compat bound for PkgX, but does want compat for all other dependencies.

lgoettgens:
Currently, only `Project.toml` and `test/Project.toml` are checked for their formatting. This PR adds the same check for `docs/Project.toml`, if that file exists.

Resolves #51.


fingolfin (approved):
Looks good to me

PR #118, opened 2023-05-31
CI: switch docs CI to standard setup

fingolfin:
This is an experiment, not sure if it will work.


fingolfin:
Seems to be working quite well :-). I've compared the resulting docs at https://juliatesting.github.io/Aqua.jl/previews/PR118/ and compared to https://juliatesting.github.io/Aqua.jl/dev/ and really couldn't see any differences

Issue #109, opened 2023-05-01
Mark tests as broken


KeithWM:
First of all: awesome package!

I think it would be nice if the keyword argument support for `test_all` also allowed marking tests as `broken`. I notice that I often discover ambiguities in dependencies. I try to fix these there, but in the mean time, I want my own CI/CD to pass, without skipping the ambiguities test altogether.

PR #124, opened 2023-06-01
Add kwarg `broken`

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/109.

This adds a `broken` kwarg (default `false`) to the tests `ambiguities`, `exports`, `piracy`, and `unbound_args`.
For the other tests it a lot harder to do as the tests are executed in a loop, and, after thinking some more about it, a lot less sensible after all.

### DIFF
--- a/src/Aqua.jl
+++ b/src/Aqua.jl
@@ -138,8 +138,8 @@ function test_all(
         end
     end
     @testset "Piracy" begin
-        if piracy
-            test_piracy(testtarget)
+        if piracy !== false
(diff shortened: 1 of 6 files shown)


fingolfin (approved):
Nice, thank you!

fingolfin:
This is marked as draft right now. Let me know when you think it is ready.

lgoettgens:
I was struggling with making the formatter happy :D

PR #128, opened 2023-06-01
Update logo

hyrodium:
This PR adds the logo to the documentation page, and updates the `Makefile`.

x-ref: https://github.com/JuliaTesting/Aqua.jl/issues/119#issuecomment-1572153748

image

image


Makefile
```diff
@@ -1,2 +1,12 @@
...
+create_badge:
+	SVG_BASE64=$(shell base64 -w 0 docs/src/assets/logo.svg); \
+	curl -o "badge.svg" "https://img.shields.io/badge/tested_with-Aqua.jl-05C3DD.svg?logo=data:image/svg+xml;base64,$$SVG_BASE64"
+
+create_favicon:
```
fingolfin, line 5: Actually this target could also be turned into a real one, no?
```suggestion
favicon.ico: docs/src/assets/logo.svg
```

hyrodium:
Hmm, the logo with the white stroke on the documenter-light mode does not look good to me. Maybe, a transparent stroke version would be better?

badge

image

image

fingolfin:
nice!!

Makefile
hyrodium: Sorry, I'm not much familiar with Makefile. Could you explain the suggestion in more detail?

fingolfin: The main purpose of `make` is to define rules how to create certain files (the "targets") from other files (the "dependencies" of the target) using certain rules. Then `make` uses file modification times to decide whether to regenerate the "target": if the target is missing, or if any of the dependencies is newer than the target, then it will run the rule to (re-)generate the target.

This is why my change proposes: it defines a target `facicon.ico` with a single dependency.

In contrast, the targets you added so far in `Makefile` are what is called "phony" target: e.g. the target `create_favicon` has no dependencies, and is simple meant to always be run when requested. When doing there, there really isn't much need for a `Makefile`, as then one could simply write a script which does the same job. (Phony targets can still be useful to support other regular targets without having to add additional script files, but here we are *only* defining phony targets.

In the end it is not super important, it just seems odd to not do this case with a "regular" target.

(BTW for `badge.svg` I did not propose a regular target; the reason is that it of course depends on whatever shields.io does, i.e., they might one day change how things look, and we can't detect that by inspecting a file locally).

Makefile
```diff
@@ -1,2 +1,16 @@
...
+all: generate_badge generate_favicon
+
+.PHONY: generate_badge generate_favicon
```
fingolfin, line 3:
```suggestion
.PHONY: all generate_badge generate_favicon
```
The target `all` is also phony because it does not actually produce a file named `all`. 

(Why declare target as phony? well, if there ever *is* a file `all` added by the user, then `make` or `make all` won't do anything anymore because `all` is already present.

fingolfin:
I personally prefer the non-transparent of the logo slightly, but in the end, I am fine either way. Perhaps @lgoettgens has a preference? Anyway, I am also happy to merge-as is, as I think this is an improvement either way. Thank you!

Makefile
hyrodium: Thank you for the detailed explanation! I will update the PR tomorrow.

hyrodium:
I made five more variations for the stroke colors.
Which color would be best?

# Aqua blue

badge

image

image

# Julia blue

badge

image

image

# Same white color as in the documenter background

badge

image

image

# No stroke

badge

image

image

# Another aqua color (`#ACDBDA`)

badge

image

image

Makefile
hyrodium: I have updated `Makefile` as suggested, and also revert `create_badge` to `badge.svg`. I agree that the target depends on what shields.io does, but I think we don't need to check the update from shields.io every time. (We can check it via `make -B` if we need to update all generated files if we want.)

Makefile
hyrodium: Thank you for the comment! I have updated the file in https://github.com/JuliaTesting/Aqua.jl/pull/128/commits/d45445733e3776fc397a4a077c1c6c4d5bc7439a.

Makefile
fingolfin: agreed

fingolfin:
I've played a bit more with the SVG, trying to make it work with a double outline. 

### With two equally thick outlines:

aqua-logo-double-outline-thick-light

aqua-logo-double-outline-thick-dark

### With a thick white and a thin black outline:

aqua-logo-double-outline-thin-light

aqua-logo-double-outline-thin-dark

hyrodium:
I feel it's better without double outline.🤓

gdalle:
I like no stroke or documenter white

lgoettgens:
This needs a rebase.

And I really have no real preference, but like the "documenter white" one a bit more

hyrodium:
Okay, I'll update this PR with documenter-white. :+1:

lgoettgens:
Please have a look at the code-style CI job and try to make it happy :)

hyrodium:
I think the format suggestion from the bot can be improved like in ChainRulesCore.jl. (https://github.com/JuliaDiff/ChainRulesCore.jl/pull/626#discussion_r1284265479)

Can I open a PR for that?

lgoettgens:
> I think the format suggestion from the bot can be improved like in ChainRulesCore.jl. ([JuliaDiff/ChainRulesCore.jl#626 (comment)](https://github.com/JuliaDiff/ChainRulesCore.jl/pull/626#discussion_r1284265479))
> 
> Can I open a PR for that?

Yeah, of course. It would be great to have that.

lgoettgens (commented):
The look is fine. Some technical comments

lgoettgens (approved):

hyrodium:
Thank you for the review & merge!

lgoettgens:
Thank You for the effort in this PR!

PR #129, opened 2023-06-01
Drop claims that 0.4 and 0.5 are still maintained

fingolfin:
At least I have no intention to maintain them; they only differ by
defaults, so anyone using Aqua can easily adapt to them if desired. In
contrast it requires additional effort to maintain three release series,
with unclear benefits.


hyrodium (approved):
LGTM!

lgoettgens (approved):

PR #130, opened 2023-06-01
Make kwarg `ignore` for `test_deps_compat` work

lgoettgens:
Currently, the kwarg `ignore` is only used to restrict the list of packages whose `Project.toml` files are investigated. This is neither sensible (as the list of these packages is user input, and most times only one package) nor matching the docstring.

This now changes it in a way, that all packages from `ignore` are removed from the list of deps with missing `[compat]` entry before concluding the test.

This came to my mind in https://github.com/thofma/Hecke.jl/pull/1086, to be precise https://github.com/thofma/Hecke.jl/pull/1086/commits/c6a42e9a05f2f3da98b4d0de6b25170fcb9fb673 adds the ignore, but it still fails in https://github.com/thofma/Hecke.jl/actions/runs/5146185098/jobs/9264865856?pr=1086#step:7:204.

Thus, it would be great to have this change in a release sometime soon-ish.


fingolfin:
Thanks!

devmotion:
Would it be possible to tag a new release with this fix? I'd like to add Aqua-tests to a package but that requires this fix :slightly_smiling_face:

fingolfin:
Once PR #131 is done and merged I'll make another release

andreasnoack:
> Once PR https://github.com/JuliaTesting/Aqua.jl/pull/131 is done and merged I'll make another release

Anything holding back the release at this point?

fingolfin:
Some other minor work is going on. I think we'll have a release today, if not then tomorrow.

PR #131, opened 2023-06-04
Fix some incorrect piracy flagging

### LINKED ISSUE #114: Incorrect (?) type piracy report
Consider this MWE:
```julia
struct MyType
  x::String
end
const MyUnion = Union{Int,MyType}
MyUnion(x::MyUnion) = x
```
Then Aqua flags the method in the final line as committing type piracy. But I don't think that's actually correct (or is it?). Note that this is *not* the same as doing, say
```
Int(x::MyUnion) = x
MyType(x::MyUnion) = x
```
In particular, the initial code snippet does provide methods for `Int(x::Int)` nor for `MyType(x::Int)`. It is a method solely for the type union `MyUnion`, which involves the type `MyType` from my package. As such, there seems to be no type piracy?

fingolfin:
If the type of a method *argument* is a union, then if *any* of the types in the union is foreign, the whole type should count as foreign.

But when installing a method for a union, then it should only count as foreign if *all* the types in the union are foreign.

Resolves #114

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -110,10 +110,36 @@ is_foreign(@nospecialize(T::TypeVar), pkg::Base.PkgId) = is_foreign(T.ub, pkg)
 end
 
 function is_foreign(@nospecialize(U::Union), pkg::Base.PkgId)
-    # Even if Foo is local, overloading f(::Union{Foo, Int}) with foreign f 
+    # Even if Foo is local, overloading f(::Union{Foo, Int}) with foreign f
     # is piracy.
     any(T -> is_foreign(T, pkg), Base.uniontypes(U))
 end
+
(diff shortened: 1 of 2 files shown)


src/piracy.jl
```diff
@@ -123,6 +149,10 @@ function is_pirate(meth::Method)
         signature = signature.body
     end
 
+    # the first parameter in the signature is the function type, and it
+    # follows slightly other rules if it happens to be a Union type
+    is_foreign_method(signature.parameters[1], method_pkg) || return false
+
     all(param -> is_foreign(param, method_pkg), signature.parameters)
```
lgoettgens, line 156: As the first entry of `signature.parameters` is handled differently, the `is_foreign` check should only be applied to `signature.parameters[2:end]`, right?

PR #134, opened 2023-06-05
Remove support for 0.4.x & 0.5.x series

fingolfin:
The only difference are the default values for certain keyword
arguments. Anyone using Aqua can easily pass different values
for these without the need of additional Aqua versions.

### DIFF
--- a/src/Aqua.jl
+++ b/src/Aqua.jl
@@ -65,22 +65,22 @@ passed to `\$x` to specify the keyword arguments for `test_\$x`.
 - `ambiguities = true`
 - `unbound_args = true`
 - `undefined_exports = true`
-- `piracy = true`
 - `project_extras = true`
 - `stale_deps = true`
 - `deps_compat = true`
 - `project_toml_formatting = true`
+- `piracy = true`
 """
 function test_all(
     testtarget::Module;
     ambiguities = true,
     unbound_args = true,
     undefined_exports = true,
-    piracy = true,
     project_extras = true,
     stale_deps = true,
     deps_compat = true,
     project_toml_formatting = true,
+    piracy = true,
 )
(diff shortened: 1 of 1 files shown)


src/Aqua.jl
```diff
@@ -77,33 +65,31 @@ passed to `\$x` to specify the keyword arguments for `test_\$x`.
 - `ambiguities = true`
 - `unbound_args = true`
 - `undefined_exports = true`
-- `piracy = $(_lt06("false", "true"))`
-- `project_extras = $(_lt05("false", "true"))`
-- `stale_deps = $(_lt05("false", "true"))`
-- `deps_compat = $(_lt05("false", "true"))`
-- `project_toml_formatting = $(_lt05("false", "true"))`
+- `piracy = true`
```
lgoettgens, line 68: Could you update the order here s.t. it is equal to the execution order. E.g. piracy should be the last.

src/Aqua.jl
```diff
@@ -77,33 +65,31 @@ passed to `\$x` to specify the keyword arguments for `test_\$x`.
...
     testtarget::Module;
     ambiguities = true,
     unbound_args = true,
     undefined_exports = true,
-    piracy = AQUA_VERSION >= v"0.6-",
-    project_extras = AQUA_VERSION >= v"0.5-",
-    stale_deps = AQUA_VERSION >= v"0.5-",
-    deps_compat = AQUA_VERSION >= v"0.5-",
-    project_toml_formatting = AQUA_VERSION >= v"0.5-",
+    piracy = true,
```
lgoettgens, line 79: Here as well

src/Aqua.jl
fingolfin: Done

Issue #107, opened 2023-04-24
Don't report dependencies as stale if they are also in `[weakdeps]` (in Julia 1.9)


gdalle:
As per the Pkg documentation for Julia 1.9, [it is valid](https://pkgdocs.julialang.org/dev/creating-packages/#Transition-from-normal-dependency-to-extension) to put some packages necessary for extensions both in `[deps]` and in `[weakdeps]`, and they don't have to be imported. At the moment, Aqua flags this behavior as a stale dep

oschulz:
Just ran into this one too.

gdalle:
But we would also want to check for actual stale deps in the extensions too, right? So I guess the criterion is rather "is the dependency imported at all, whether in the main package or in an extension"

fingolfin:
@gdalle what you describe there would also be nice to have, but it's a step beyond what this issue requests: to avoid incorrect stale dependency reports, i.e., false positives. Which should be fairly easy to implement.

Once we have that, of course it would be nice if the stale dependency checker was further improve to deal with extensions, to avoid false negatives.

lgoettgens:
> I think this wouldn't be too hard to fix (that doesn't mean I'll get to it anytime soon; patches welcome). I'd start by looking at `_analyze_stale_deps_1` which obtains the list of `deps`. I'd change it to also obtain the `weakdeps`, and then remove anything from `deps` that is also in `weakdeps`. That should be enough.

Working on this. Expect a PR this afternoon.

PR #135, opened 2023-06-05
Do not report stale deps that are weakdeps as well

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/107.


fingolfin:
Thank you!

Issue #138, opened 2023-06-16
TypePiracy: treat foreign type as own type


lgoettgens:
Originally brought up in https://github.com/oscar-system/GAP.jl/pull/889#issuecomment-1578075249. 

The julia docu explains some scenario, where one would have a good reason to do type piracy, e. g. when splitting a C wrapper into a lightweight jll defining the types and another julia package implementing higher level functionality, see https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy-1 (last paragraph).

I would thus propose to add a kwarg to the piracy test with a list of types that should be treated as own types even if they are foreign.
Most other Aqua test cases have some kind of `ignore` kwarg, so I see no point in not having something similar for `test_piracy`.

devmotion:
It would also be useful to be able to specify that a function is owned by a different package: For instance, StatsAPI defines function stumps that are officially declared to be owned by StatsBase - but currently Aqua flags their default implementation for generic types (defined in Base) as type piracy (https://github.com/JuliaStats/StatsBase.jl/pull/866).

lgoettgens:
So a solution to you would be to split up my proposed kwarg `treat_as_own` into two different ones: one for types that works as proposed in #140, and one for functions, which just doesn't check type piracy for the functions given there.
Is that what you have in mind?

And I would like to hear the opinion of @fingolfin on this - If there is nothing I have missed, I will update my pr #140 to include your idea as well.

devmotion:
Yes, I think such a generalization would fix my use case in StatsBase as well.

Issue #96, opened 2023-01-23
Allow type piracy from superpackage?


kshyatt:
Suppose I have two packages `A` and `B`, and `B` is a subpackage of `A` (they live in the same repo), perhaps to provide additional, optional functionality (e.g. GPU integration). In such a case, I might actually want `B` to engage in some type piracy from `A` - is there a way to allow that in the `Aqua` checks for `B`?

jakobnissen:
Not yet, but this would be a good idea. Perhaps to provide a set of packages from which your package is allowed to pirate.

PR #140, opened 2023-06-16
Piracy: Add kwarg `treat_as_own`

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/138, resolves https://github.com/JuliaTesting/Aqua.jl/issues/96.

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -47,8 +47,8 @@ end
 ##################################
 # Generic fallback for type parameters that are instances, like the 1 in
 # Array{T, 1}
-is_foreign(@nospecialize(x), pkg::Base.PkgId; treat_as_own::Vector{<:Type} = Type[]) =
-    is_foreign(typeof(x), pkg; treat_as_own = treat_as_own)
+is_foreign(@nospecialize(x), pkg::Base.PkgId; treat_as_own_type::Vector{<:Type} = Type[]) =
+    is_foreign(typeof(x), pkg; treat_as_own_type = treat_as_own_type)
 
 # Symbols can be used as type params - we assume these are unique and not
 # piracy.  This implies that we have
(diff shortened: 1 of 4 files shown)


lgoettgens:
This now contains two `treat_as_own` kwargs, one for types and one for functions (cf. https://github.com/JuliaTesting/Aqua.jl/issues/138#issuecomment-1596275081).

This is still a draft as I am unhappy with the names `treat_as_own_type` and `treat_as_own_func`.
Maybe @devmotion or @fingolfin or someone else has an idea?

src/piracy.jl
```diff
@@ -77,71 +78,118 @@ is_foreign(@nospecialize(x), pkg::Base.PkgId) = is_foreign(typeof(x), pkg)
 # a crazy API).  The symbol name may also come from `gensym`.  Since the aim of
 # `Aqua.test_piracy` is to detect only "obvious" piracy, let us play on the
 # safe side.
-is_foreign(x::Symbol, pkg::Base.PkgId) = false
+is_foreign(x::Symbol, pkg::Base.PkgId; treat_as_own_type::Vector{<:Type} = Type[]) = false
```
fingolfin, line 81: Do we ever *not* pass these `treat_as_own`? If not, then why even provide a default value?

Also, just wondering: why use `Vector` and not `Set`? For shorter (or empty) lists it may not matter, but perhaps if they are bigger? For that matter, we could simply not restrict the type here at all?)
```suggestion
is_foreign(x::Symbol, pkg::Base.PkgId, treat_as_own_type::Set)
```

src/piracy.jl
```diff
@@ -77,71 +78,118 @@ is_foreign(@nospecialize(x), pkg::Base.PkgId) = is_foreign(typeof(x), pkg)
...
-    any(T -> is_foreign(T, pkg), Base.uniontypes(U))
+    any(T -> is_foreign(T, pkg; treat_as_own_type = treat_as_own_type), Base.uniontypes(U))
 end
 
-function is_foreign_method(@nospecialize(U::Union), pkg::Base.PkgId)
+function is_foreign_method(
+    @nospecialize(U::Union),
+    pkg::Base.PkgId;
+    treat_as_own_type::Vector{<:Type} = Type[],
+    treat_as_own_func::Vector{<:Function} = Function[],
```
fingolfin, line 146: What do we gain by having two arguments? It seems fine to mix both types and functions in a single list, and arguably might be easier to use and read?

src/piracy.jl
```diff
@@ -173,9 +230,16 @@ See [Julia documentation](https://docs.julialang.org/en/v1/manual/style-guide/#A
 # Keyword Arguments
 - `broken::Bool = false`: If true, it uses `@test_broken` instead of
   `@test`.
+- `treat_as_own_type::Vector{<:Type} = Type[]`: The types in this vector 
+  are considered to be "owned" by the module `m`. This is useful for 
+  testing packages that deliberately commit some type piracy, e.g. modules 
+  adding higher-level functionality to a lightweight C-wrapper.
+- `treat_as_own_func::Vector{<:Function} = Function[]`: The functions in 
+  this vector are considered to be "owned" by the module `m`. This is useful,
+  e. g. for testing packages that are extending StatsAPI.jl.
```
fingolfin, line 239:
```suggestion
- `treat_as_own = Type[]`: The types and functions in this container 
  are considered to be "owned" by the module `m`. This is useful for 
  testing packages that deliberately commit some type piracy, e.g. packages 
  adding higher-level functionality to a lightweight C-wrapper, or
  packages that are extending StatsAPI.jl.
```

fingolfin (commented):
Unfortunately this now needs conflict resolution.

src/piracy.jl
lgoettgens: I guess the only downside to this is that julia cannot specialize in the parameters anymore. But this should not make that big of a difference. Thus I merged the two kwargs in 9e0a05b7841d5fa00585b93828682b810f55f114.

fingolfin:
Oops another conflict 😞

lgoettgens:
Thanks for resolving!

Issue #141, opened 2023-06-18
Ambiguity check errors on Julia 1.0: `type UnionAll has no field name`


devmotion:
In https://github.com/JuliaStats/StatsBase.jl/pull/866, I observed test errors of the ambiguity check on Julia 1.0 which seem to be caused by a bug in Aqua (see, e.g., https://github.com/JuliaStats/StatsBase.jl/actions/runs/5298486934/jobs/9590870427?pr=866):
```julia
┌ Error: Failed to obtain a function from `Method`.
│   exception =
│    type UnionAll has no field name
│    Stacktrace:
│     [1] getproperty at ./sysimg.jl:15 [inlined]
│     [2] getobj(::Method) at /home/runner/.julia/packages/Aqua/BhBic/src/ambiguities.jl:168
...
```

fingolfin:
Thank you for the report. I think this is a duplicate of issue #49.

Issue #49, opened 2020-09-30
`test_ambiguities` errors when `exclude` is set.


mtsch:
Hi,
Thanks for the great package! The new version of Aqua appears to have problems with the `exclude` argument in `test_ambiguities`. I get this error on Julia 1.0 and 1.5, but not on nightly. [Here is an example Travis log](https://travis-ci.com/github/mtsch/Ripserer.jl/builds/187363831).

For an MWE, create a package MWE.jl with:

`/path/to/MWE.jl/src/mwe.jl`:

```
module MWE
using StaticArrays
end
```

Then, in a Julia session:
```
julia> using MWE
julia> using Aqua
julia> Aqua.test_ambiguities([MWE, Base], exclude=[convert])
# passes
julia> Aqua.test_ambiguities([MWE, Base, Core])
# fails normally
julia> Aqua.test_ambiguities([MWE, Base, Core], exclude=[convert])
Skipping Base.Filesystem.JL_O_RANDOM
Skipping Base.Filesystem.JL_O_SEQUENTIAL
Skipping Base.Filesystem.JL_O_SHORT_LIVED
Skipping Base.Filesystem.JL_O_TEMPORARY
Skipping Base.active_repl
Skipping Base.active_repl_backend
┌ Error: Failed to obtain a function from `Method`.
│   exception =
│    type UnionAll has no field name
│    Stacktrace:
│     [1] getproperty at ./Base.jl:28 [inlined]
│     [2] getobj(::Method) at /home/m/.julia/packages/Aqua/oqyfA/src/ambiguities.jl:161
│     [3] (::Aqua.var"#16#17"{Array{typeof(convert),1}})(::Tuple{Method,Method}) at /home/m/.julia/packages/Aqua/oqyfA/src/ambiguities.jl:186
│     [4] filter(::Aqua.var"#16#17"{Array{typeof(convert),1}}, ::Array{Tuple{Method,Method},1}) at ./array.jl:2462
│     [5] test_ambiguities_impl(::Array{Base.PkgId,1}, ::NamedTuple{(:recursive,),Tuple{Bool}}, ::Array{Pair{Base.PkgId,String},1}) at /home/m/.julia/packages/Aqua/oqyfA/src/ambiguities.jl:184
│     [6] top-level scope at none:8
│     [7] eval(::Module, ::Any) at ./boot.jl:331
│     [8] exec_options(::Base.JLOptions) at ./client.jl:272
[...]

PR #142, opened 2023-06-19
Fix `getobj` for parameterized callable objects

lgoettgens:
Fixes https://github.com/JuliaTesting/Aqua.jl/issues/49 and fixes https://github.com/JuliaTesting/Aqua.jl/issues/141.

The main change is the unwrapping of `UnionAll`s in `getobj`. While looking through it, I furthermore used it in the piracy tests as well (where until now the function was basically inlined). And in `getobj` the deprecated function `tuple_type_head` was used, the alternative is `fieldtype(-, 1)`. Both changes have been tested in julia 1.0, 1.6 and 1.9.

The MWE in https://github.com/JuliaTesting/Aqua.jl/issues/49#issue-711726567 by @mtsch for julia 1.0 (`Aqua.test_ambiguities([MWE, Base, Core], exclude=[convert])`) reports precisely the same 52 ambiguities (up to order) as with the current master, but there are no errors. 

The much smaller MWE in https://github.com/JuliaTesting/Aqua.jl/issues/49#issuecomment-1576304828 by @fingolfin  for julia 1.9 no longer fails.

Please check if this fixes it for you as well.

### DIFF
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -150,7 +150,7 @@ struct _NoValue end
 # FIXME: This does not support non-singleton callables.
 function getobj(m::Method)
     ty = try
-        Base.tuple_type_head(m.sig)
+        fieldtype(m.sig, 1)
     catch err
         @error(
             "Failed to obtain a function from `Method`.",
@@ -160,6 +160,7 @@ function getobj(m::Method)
         # too much compared to what it is now.  So, bailing out.
         return _NoValue()
     end
+    ty = Base.unwrap_unionall(ty)
     try
         return ty.instance  # this should work for singletons (e.g., functions)
     catch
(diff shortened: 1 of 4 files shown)


lgoettgens:
For the added testcase in the style of the method erroring https://github.com/JuliaTesting/Aqua.jl/issues/49#issuecomment-1576304828, one can now see the error in the CI logs (e.g. https://github.com/JuliaTesting/Aqua.jl/actions/runs/5313233395/jobs/9618834173?pr=142#step:4:139). Unfortunately, due to the implementation of `test_ambiguities`, we are not able to test for their absence. I'll try to circumvent that somehow.

fingolfin (approved):
Looks good to me, thanks!

PR #143, opened 2023-06-20
Refactor `ambiguities.jl`

lgoettgens:
Changes:

- Formatting, it annoys me every time I want to change something that I cannot run the autoformatter
- Parse the output of `test_ambiguities_impl` to allow better testing (number of ambiguities and errors inside `getobj`)
- Adapt the tests

### DIFF
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -85,11 +85,23 @@ function reprexclude(exspecs::Vector{ExcludeSpec})
     return string("Aqua.ExcludeSpec[", join(itemreprs, ", "), "]")
 end
 
-function _test_ambiguities(
+function _test_ambiguities(packages::Vector{PkgId}; broken::Bool = false, kwargs...)
+    num_ambiguities, strout, strerr = _find_ambiguities(packages; kwargs...)
+
+    println(stderr, strerr)
+    println(stdout, strout)
+
+    if broken
+        @test_broken num_ambiguities == 0
+    else
+        @test num_ambiguities == 0
+    end
+end
+
+function _find_ambiguities(
     packages::Vector{PkgId};
     color::Union{Bool,Nothing} = nothing,
     exclude::AbstractVector = [],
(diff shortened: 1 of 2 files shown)


devmotion:
> Formatting, it annoys me every time I want to change something that I cannot run the autoformatter

I guess the style should be defined in a `.JuliaFormatter.toml` file?

lgoettgens:
> > Formatting, it annoys me every time I want to change something that I cannot run the autoformatter
> 
> I guess the style should be defined in a `.JuliaFormatter.toml` file?

This project is mostly using the default style

devmotion:
Which default style? IIRC, e.g., VSCode requires a `.JuliaFormatter.toml` and does not support any global or default options. Generally, it also seems safer to fix a style instead of relying on local settings of developers.

lgoettgens:
> Which default style? IIRC, e.g., VSCode requires a `.JuliaFormatter.toml` and does not support any global or default options. Generally, it also seems safer to fix a style instead of relying on local settings of developers.

I am using VSCode with an empty `.JuliaFormatter.toml`. This then falls back to some defaults defined in `JuliaFormatter.jl`.
But I agree with you, that we should add a `.JuliaFormatter.toml` to `Aqua`, but not in this PR.

devmotion:
> I am using VSCode with an empty .JuliaFormatter.toml

But that's your setting and someone else might use different options :slightly_smiling_face: 

> But I agree with you, that we should add a .JuliaFormatter.toml to Aqua, but not in this PR.

IMO formatting changes due to the autoformatter are better motivated if the style is clearly fixed. Otherwise someone else might open another PR with formatting changes based on different style options :slightly_smiling_face:

fingolfin:
This now needs a rebase.

lgoettgens:
This is now rebased. The formatting commit now only contains two minor changes, where I made the code a bit more readable and shorter compared to the formatter.

PR #144, opened 2023-06-20
Enhance `test_ambiguities`

lgoettgens:
As written in the removed comment, `test_ambiguities` does not allow non-singleton callable in `exclude`. This is working now.
I furthermore changed the behavior for singleton callables to be the same. And I introduced constructors for `exclude`.
Unfortunately, one can only exclude the constructor of type `MyType` and all callables of `MyType` simultaneously, or none of both.

For my convenience, this already includes the changes of https://github.com/JuliaTesting/Aqua.jl/pull/143.

This change is probably breaking, semi-documented behavior of `exclude` for `test_ambiguities`. I thus bumped the version.


fingolfin:
Needs a rebase

PR #145, opened 2023-06-20
Add `.JuliaFormatter.toml`

lgoettgens:
I would like to fix some style. I do not care (in most cases), how exactly it looks (thus defaults).

cc @devmotion

### DIFF
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -26,25 +26,23 @@ false-positive.
 - Other keyword arguments such as `imported` and `ambiguous_bottom`
   are passed to `Test.detect_ambiguities` as-is.
 """
-test_ambiguities(packages; kwargs...) =
-    _test_ambiguities(aspkgids(packages); kwargs...)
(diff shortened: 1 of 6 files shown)


fingolfin:
There is a code formatting CI job, so the settings must match with that

lgoettgens:
> There is a code formatting CI job, so the settings must match with that

I reformatted all files with the empty `.JuliaFormatter.toml`. If the formatting CI job succeeds, there seems to be no reason to not adopt the config as is.

Yes, this will bring up a lot of conflicts with other PRs. But since those are mainly by me, I am willing to sort them out.

devmotion:
Base on the format action, there seems no reason for reformatting the files: It succeeded also on the first commit which only added `.JuliaFormatter.toml` (https://github.com/JuliaTesting/Aqua.jl/actions/runs/5326268396/jobs/9648076565).

Or, maybe that indicates that the action does not actually detect incorrectly formatted code?

PR #149, opened 2023-06-24
Remove special branch for some julia v1.6.0-DEV versions

lgoettgens:
This branch contains some special code for versions between https://github.com/JuliaLang/julia/pull/36962 (`v1.6.0-DEV.816`) and https://github.com/JuliaLang/julia/pull/37484 (`v1.6.0-DEV.875`).

I think Aqua should not need to support DEV builds that are more than 2 years old.
Thus I propose to remove the branch.

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -108,7 +108,7 @@ is_foreign(@nospecialize(T::TypeVar), pkg::Base.PkgId; treat_as_own) =
     is_foreign(T.ub, pkg; treat_as_own = treat_as_own)
 
 # Before 1.7, Vararg was a UnionAll, so the UnionAll method will work
-@static if VERSION >= v"1.7"
+@static if VERSION >= v"1.7-"
     is_foreign(@nospecialize(T::Core.TypeofVararg), pkg::Base.PkgId; treat_as_own) =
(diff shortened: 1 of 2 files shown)


src/piracy.jl
```diff
@@ -105,7 +105,7 @@ end
 is_foreign(@nospecialize(T::TypeVar), pkg::Base.PkgId) = is_foreign(T.ub, pkg)
 
 # Before 1.7, Vararg was a UnionAll, so the UnionAll method will work
-@static if VERSION >= v"1.7"
+@static if VERSION >= v"1.7-"
```
fingolfin, line 108: Huh, what is the difference in behavior herr?

lgoettgens: the new comparison is also true for all beta's and rc's before the 1.7 release. As the behavior tested here was changed before the 1.7 feature freeze, this should be changed, as it currently breaks tests on e.g. `1.7.0-beta1` and `1.7.0-rc3`. Changing this fixes it.

fingolfin: OK, so this is for a hypothetical scenario were someone is *today* using a beta version of an outdated non-LTS Julia version... got it ;-)

PR #152, opened 2023-06-26
Revert "Enhance `test_ambiguities`"

lgoettgens:
Reverts JuliaTesting/Aqua.jl#144.

For reasoning see https://github.com/JuliaTesting/Aqua.jl/pull/153.


fingolfin:
Not needed

PR #156, opened 2023-06-29
Enhance Piracy detection by considering more methods

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/155 and resolves https://github.com/JuliaTesting/Aqua.jl/issues/157.

The main change is a re-implementation of `all_methods`. Unfortunately, the old data source `names(MyModule)` was missing some callables (https://github.com/JuliaTesting/Aqua.jl/issues/157) and methods of non-imported functions that have been defined in a fully qualified way (https://github.com/JuliaTesting/Aqua.jl/issues/155). The proposed implementation works very similarly to `Test.detect_ambiguities` and `Test.detect_unbound_args` by iterating over all `names` of all currently loaded modules, and then filtering out the methods that are indeed defined in `MyModule`.

I furthermore added some tests displaying the new functionality. The main point here is that https://github.com/JuliaTesting/Aqua.jl/blob/94a9c7a69ee41878774ce2dc4cac05fb4f71c4ed/test/test_piracy.jl#L58-L62
is no longer needed.

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -1,6 +1,18 @@
 module Piracy
 
-import Test
+if VERSION >= v"1.6-"
+    using Test: is_in_mods
+else
+    function is_in_mods(m::Module, recursive::Bool, mods)
+        while true
+            m in mods && return true
+            recursive || return false
+            p = parentmodule(m)
+            p === m && return false
+            m = p
+        end
+    end
+end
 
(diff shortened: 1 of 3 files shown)


fingolfin (approved):
I have no idea what this does exactly but I trust you thought this through.

Issue #165, opened 2023-08-24
Regression in ambiguities test


fingolfin:
I just tried to reproduce the list of ambiguities for AbstractAlgebra.jl, via
```
julia> using Aqua, AbstractAlgebra; Aqua.test_all(
           AbstractAlgebra;
           ambiguities=true,
           unbound_args=true,
           undefined_exports=true,
           project_extras=true,
           stale_deps=true,
           deps_compat=true,
           project_toml_formatting=true,
           piracy=false            # TODO: fix piracy
       )
```
Unfortunately, this just hangs in the ambiguities test. However, it works when I downgrade Aqua.jl to 0.6.3.

Using `git bisect` I determined that this regression was introduced in commit e391873efc31425f8968f136e0e19fc447a8940e from PR #143 by @lgoettgens.

fingolfin:
Some more information: this is with Julia 1.9.3 on macOS on ARM.

If I interrupt via ctrl-C, this is the stacktrace I see:
```
^CTest Summary:    |Time
Method ambiguity | None  5.3s
ERROR: InterruptException:
Stacktrace:
  [1] poptask(W::Base.IntrusiveLinkedListSynchronized{Task})
    @ Base ./task.jl:974
  [2] wait()
    @ Base ./task.jl:983
  [3] wait(c::Base.GenericCondition{Base.Threads.SpinLock}; first::Bool)
    @ Base ./condition.jl:130
  [4] wait
    @ ./condition.jl:125 [inlined]
  [5] wait(x::Base.Process)
    @ Base ./process.jl:661
  [6] success(x::Base.Process)
    @ Base ./process.jl:523
  [7] success(cmd::Base.CmdRedirect)
    @ Base ./process.jl:536
  [8] _find_ambiguities(packages::Vector{Base.PkgId}; color::Nothing, exclude::Vector{Any}, detect_ambiguities_options::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Aqua ~/Projekte/Julia/packages/JuliaTesting/Aqua.jl/src/ambiguities.jl:130
  [9] _find_ambiguities
    @ ~/Projekte/Julia/packages/JuliaTesting/Aqua.jl/src/ambiguities.jl:100 [inlined]
[...]

PR #166, opened 2023-08-24
Circumvent pipe buffer deadlock

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/165.

@fingolfin please check that on your machine as well.

This should be backported to 0.6.6 and released, as it should occur for every package with many (>150 ambiguities).

### DIFF
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -126,7 +126,7 @@ function _find_ambiguities(
     end
     cmd = `$cmd --startup-file=no -e $code`
 
-    outfile, out = mktemp(; cleanup = false)
+    outfile, out = mktemp()
     errfile, err = mktemp(; cleanup = false)
     succ = success(pipeline(cmd; stdout = out, stderr = err))
     strout = read(outfile, String)
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -126,13 +126,11 @@ function _find_ambiguities(
     end
     cmd = `$cmd --startup-file=no -e $code`
 
-    out = Pipe()
-    err = Pipe()
+    outfile, out = mktemp(; cleanup = false)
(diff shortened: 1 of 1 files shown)


fingolfin (approved):

src/ambiguities.jl
```diff
@@ -126,13 +126,11 @@ function _find_ambiguities(
     end
     cmd = `$cmd --startup-file=no -e $code`
 
-    out = Pipe()
-    err = Pipe()
+    outfile, out = mktemp(; cleanup = false)
```
fingolfin, line 129: Unfortunately this fails in Julia 1.0 and 1.1:
```
  function mktemp does not accept keyword arguments
```

src/ambiguities.jl
```diff
@@ -126,13 +126,11 @@ function _find_ambiguities(
     end
     cmd = `$cmd --startup-file=no -e $code`
 
-    out = Pipe()
-    err = Pipe()
+    outfile, out = mktemp()
+    errfile, err = mktemp(; cleanup = false)
```
lgoettgens, line 130:
```suggestion
    errfile, err = mktemp()
```

src/ambiguities.jl
```diff
@@ -126,13 +126,11 @@ function _find_ambiguities(
...
-    err = Pipe()
+    outfile, out = mktemp()
+    errfile, err = mktemp()
     succ = success(pipeline(cmd; stdout = out, stderr = err))
-    close(out.in)
-    close(err.in)
-    strout = String(read(out))
-    strerr = String(read(err))
+    strout = read(outfile, String)
+    strerr = read(errfile, String)
```
fingolfin, line 133: should we `rm(outfile)` and `rm(errfile)`?

PR #171, opened 2023-09-04
Piracy fix for kwcall

lgoettgens:
The implementation of https://github.com/JuliaTesting/Aqua.jl/pull/131 does not work properly if the function in question has a kwarg (see the change in the test file here for an example). 
This fixes that by ignoring the first two elements of the signature of a `kwcall` (first is `typeof(kwcall)` and second is `Any`, the NamedTuple of all kwargs to be applied), and then considering the third to be "the original function" and everything behind that as positional arguments.

### DIFF
--- a/src/piracy.jl
+++ b/src/piracy.jl
@@ -162,14 +162,25 @@ function is_pirate(meth::Method; treat_as_own = Union{Function,Type}[])
 
     signature = Base.unwrap_unionall(meth.sig)
 
+    function_type_index = 1
+    if signature.parameters[1] === typeof(Core.kwcall)
+        # kwcall is a special case, since it is not a real function
+        # but a wrapper around a function, the third parameter is the original
+        # function, its positional arguments follow.
+        function_type_index += 2
+    end
+
     # the first parameter in the signature is the function type, and it
     # follows slightly other rules if it happens to be a Union type
-    is_foreign_method(signature.parameters[1], method_pkg; treat_as_own = treat_as_own) ||
-        return false
+    is_foreign_method(
+        signature.parameters[function_type_index],
+        method_pkg;
+        treat_as_own = treat_as_own,
+    ) || return false
 
-    all(
+    return all(
         param -> is_foreign(param, method_pkg; treat_as_own = treat_as_own),
-        signature.parameters[2:end],
+        signature.parameters[function_type_index+1:end],
     )
 end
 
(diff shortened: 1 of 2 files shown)


fingolfin (approved):

fingolfin (changes requested):
Unfortunately it fails in all older Julia versions (<=1.8)

lgoettgens:
> Unfortunately it fails in all older Julia versions (<=1.8)

Works now in julia 1.5. Let's see if CI succeeds for older version as well. It needed way much more of wiggling around with internals than I expected or wanted.

src/utils.jl
```diff
@@ -157,3 +157,22 @@ function format_diff(
     $text_b
     """
 end
+
+function is_kwcall(signature::DataType)
```
fingolfin, line 161: Perhaps it would make sense to use https://github.com/FluxML/MacroTools.jl or https://github.com/invenia/ExprTools.jl instead of rolling this on our own?

```julia
julia> ex = :(
                  function Base.f(x::T, y::T; z::Bool=false) where T
                      x + y
                  end
              );

julia> ExprTools.splitdef(ex)
Dict{Symbol, Any} with 6 entries:
  :args        => Any[:(x::T), :(y::T)]
  :body        => quote…
  :kwargs      => Any[:($(Expr(:kw, :(z::Bool), false)))]
  :head        => :function
  :whereparams => Any[:T]
  :name        => :(Base.f)

julia> MacroTools.splitdef(ex)
Dict{Symbol, Any} with 5 entries:
  :name        => :(Base.f)
  :args        => Any[:(x::T), :(y::T)]
  :kwargs      => Any[:($(Expr(:kw, :(z::Bool), false)))]
  :body        => quote…
  :whereparams => (:T,)
```

fingolfin: Ah, but of course that's for expressions, not methods/signatures, so it might not be helpful.

lgoettgens: Yeah, these ones unfortunately do not offer what we need here. I couldn't find any *good way* to test this in other packages. Most just ignore old julia versions, where `Core.kwcall` doesn't exist, or use some regex with the function name of the kwsorter. The first is not an option here (unless we drop support for julia-pre-1.9, or have different semantics for the tests depending on the julia version). The second option relies on weirder internals, where we won't get good error messages in future changes (as regexes and strings just tell "no", not fail with some type error etc.). Thus, I think the current proposal here is the best we can currently do. I already added a try-catch with some additional debug information for people to open issues here.

fingolfin (approved):

PR #174, opened 2023-09-10
Add persistent_tasks test

timholy:
This checks whether a package spawns persistent `Task`s that might block precompilation of dependents of the package. It also provides a convenience utility, `test_persistent_tasks_deps`, for checking all the dependencies of a given package, to simplify discovery. If one runs one of the tests in this PR without annotating one as an expected failure, here is the output on Julia 1.10-beta:

```julia
julia> Aqua.test_persistent_tasks_deps(getid("UsesBoth"))
  Generating  project jl_jqh8hMTOr6:
    /tmp/jl_jqh8hMTOr6/Project.toml
    /tmp/jl_jqh8hMTOr6/src/jl_jqh8hMTOr6.jl
  Activating project at `/tmp/jl_jqh8hMTOr6`
   Resolving package versions...
    Updating `/tmp/jl_jqh8hMTOr6/Project.toml`
  [94ae9332] + TransientTask v0.1.0 `~/.julia/dev/Aqua/test/pkgs/PersistentTasks/TransientTask`
    Updating `/tmp/jl_jqh8hMTOr6/Manifest.toml`
  [94ae9332] + TransientTask v0.1.0 `~/.julia/dev/Aqua/test/pkgs/PersistentTasks/TransientTask`
  Generating  project jl_FI7OrSlgVA:
    /tmp/jl_FI7OrSlgVA/Project.toml
    /tmp/jl_FI7OrSlgVA/src/jl_FI7OrSlgVA.jl
  Activating project at `/tmp/jl_FI7OrSlgVA`
   Resolving package versions...
    Updating `/tmp/jl_FI7OrSlgVA/Project.toml`
  [e5c298b6] + PersistentTask v0.1.0 `~/.julia/dev/Aqua/test/pkgs/PersistentTasks/PersistentTask`
    Updating `/tmp/jl_FI7OrSlgVA/Manifest.toml`
  [e5c298b6] + PersistentTask v0.1.0 `~/.julia/dev/Aqua/test/pkgs/PersistentTasks/PersistentTask`
PersistentTask [e5c298b6-d81d-47aa-a9ed-5ea983e22986] persistent_tasks: Test Failed at /home/tim/.julia/dev/Aqua/src/persistent_tasks.jl:77
  Expression: fails ⊻ precompile_wrapper(result, tmax)

Stacktrace:
 [1] macro expansion
   @ Aqua ~/.julia/juliaup/julia-1.10.0-beta2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ Aqua ~/.julia/dev/Aqua/src/persistent_tasks.jl:77 [inlined]
 [3] macro expansion
   @ Aqua ~/.julia/juliaup/julia-1.10.0-beta2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] test_persistent_tasks(package::PkgId; tmax::Int64, fails::Bool)
   @ Aqua ~/.julia/dev/Aqua/src/persistent_tasks.jl:75
Test Summary:                                                             | Pass  Fail  Total   Time
/home/tim/.julia/dev/Aqua/test/pkgs/PersistentTasks/UsesBoth/Project.toml |    1     1      2  10.2s
  TransientTask [94ae9332-58b0-4342-989c-0a7e44abcca9] persistent_tasks   |    1            1   2.5s
  PersistentTask [e5c298b6-d81d-47aa-a9ed-5ea983e22986] persistent_tasks  |          1      1   7.7s
ERROR: Some tests did not pass: 1 passed, 1 failed, 0 errored, 0 broken.
```

Closes https://github.com/JuliaLang/julia/pull/50914

CC @vtjnash

### DIFF
--- a/src/persistent_tasks.jl
+++ b/src/persistent_tasks.jl
@@ -17,8 +17,6 @@ but block precompilation of any packages that depend on it.
 Let's create a dummy package, `PkgA`, that launches a persistent `Task`:
 
 ```julia
-
-```julia
 module PkgA
 const t = Ref{Any}()   # to prevent the Timer from being garbage-collected
 __init__() = t[] = Timer(0.1; interval=1)   # create a persistent `Timer` `Task`
@@ -67,10 +65,11 @@ to launch the tasks and cleanly shut them down.
 - `package`: a top-level `Module` or `Base.PkgId`.
 
 # Keyword Arguments
-- `tmax::Real`: the maximum time (in seconds) to wait between loading the
-  package and forced shutdown of the precompilation process.
+- `tmax::Real`: the maximum time (in seconds) to wait after loading the
+  package before forcibly shutting down the precompilation process (triggering
+  a test failure).
 """
-function test_persistent_tasks(package::PkgId; tmax=5, fails::Bool=false)
+function test_persistent_tasks(package::PkgId; tmax = 10, fails::Bool = false)
     @testset "$package persistent_tasks" begin
         result = root_project_or_failed_lazytest(package)
         result isa LazyTestResult && return result
@@ -83,11 +82,14 @@ function test_persistent_tasks(package::Module; kwargs...)
 end
 
 """
-    Aqua.test_persistent_tasks_deps(package; kwargs...)
+    Aqua.test_persistent_tasks_deps(package; fails = Dict{String,Bool}(), kwargs...)
 
 Test all the dependencies of `package` with [`Aqua.test_persistent_tasks`](@ref).
+`get(fails, dep, false)` encodes whether dependency `dep` is expected to fail the test
+(this is primarily intended for use in Aqua's own internal test suite).
+Any additional kwargs (e.g., `tmax`) are passed to [`Aqua.test_persistent_tasks`](@ref).
 """
-function test_persistent_tasks_deps(package::PkgId; fails=Dict{String,Bool}(), kwargs...)
+function test_persistent_tasks_deps(package::PkgId; fails = Dict{String,Bool}(), kwargs...)
     result = root_project_or_failed_lazytest(package)
     result isa LazyTestResult && return result
     prj = TOML.parsefile(result)
@@ -98,7 +100,7 @@ function test_persistent_tasks_deps(package::PkgId; fails=Dict{String,Bool}(), k
         else
             for (name, uuid) in deps
                 id = PkgId(UUID(uuid), name)
-                test_persistent_tasks(id; fails=get(fails, name, false), kwargs...)
+                test_persistent_tasks(id; fails = get(fails, name, false), kwargs...)
             end
         end
     end
@@ -112,24 +114,28 @@ function precompile_wrapper(project, tmax)
     pkgdir = dirname(project)
     pkgname = basename(pkgdir)
     wrapperdir = tempname()
-    wrappername, wrapperuuid = only(Pkg.generate(wrapperdir))
+    wrappername, _ = only(Pkg.generate(wrapperdir))
     Pkg.activate(wrapperdir)
-    Pkg.develop(PackageSpec(path=pkgdir))
+    Pkg.develop(PackageSpec(path = pkgdir))
+    statusfile = joinpath(wrapperdir, "done.log")
     open(joinpath(wrapperdir, "src", wrappername * ".jl"), "w") do io
-        println(io, """
-        module $wrappername
-        using $pkgname
-        # Signal Aqua from the precompilation process that we've finished loading the package
-        open(joinpath("$wrapperdir", "done.log"), "w") do io
-            println(io, "done")
-        end
-        end
-        """)
+        println(
+            io,
+            """
+module $wrappername
+using $pkgname
+# Signal Aqua from the precompilation process that we've finished loading the package
+open($statusfile, "w") do io
+    println(io, "done")
+end
+end
+""",
+        )
     end
     # Precompile the wrapper package
     cmd = `$(Base.julia_cmd()) --project=$wrapperdir -e 'using Pkg; Pkg.precompile()'`
-    proc = run(cmd; wait=false)
-    while !isfile(joinpath(wrapperdir, "done.log"))
+    proc = run(cmd; wait = false)
+    while !isfile(statusfile)
         sleep(0.1)
     end
     # Check whether precompilation finishes in the required time
--- a/src/persistent_tasks.jl
+++ b/src/persistent_tasks.jl
@@ -116,9 +116,14 @@ end
 
 function precompile_wrapper(project, tmax)
     prev_project = Base.active_project()
+    isdefined(Pkg, :respect_sysimage_versions) && Pkg.respect_sysimage_versions(false)
     try
         pkgdir = dirname(project)
-        pkgname = basename(pkgdir)
+        pkgname = get(TOML.parsefile(project), "name", nothing)
+        if isnothing(pkgname)
+            @error "Unable to locate package name in $project"
+            return false
+        end
         wrapperdir = tempname()
         wrappername, _ = only(Pkg.generate(wrapperdir))
         Pkg.activate(wrapperdir)
@@ -160,6 +165,7 @@ end
         end
         return success
     finally
+        isdefined(Pkg, :respect_sysimage_versions) && Pkg.respect_sysimage_versions(true)
         Pkg.activate(prev_project)
     end
 end
--- a/src/persistent_tasks.jl
+++ b/src/persistent_tasks.jl
@@ -115,17 +115,19 @@ function find_persistent_tasks_deps(package::Module; kwargs...)
 end
 
 function precompile_wrapper(project, tmax)
-    pkgdir = dirname(project)
-    pkgname = basename(pkgdir)
-    wrapperdir = tempname()
-    wrappername, _ = only(Pkg.generate(wrapperdir))
-    Pkg.activate(wrapperdir)
-    Pkg.develop(PackageSpec(path = pkgdir))
-    statusfile = joinpath(wrapperdir, "done.log")
-    open(joinpath(wrapperdir, "src", wrappername * ".jl"), "w") do io
-        println(
-            io,
-            """
+    prev_project = Base.active_project()
+    try
+        pkgdir = dirname(project)
+        pkgname = basename(pkgdir)
+        wrapperdir = tempname()
+        wrappername, _ = only(Pkg.generate(wrapperdir))
+        Pkg.activate(wrapperdir)
+        Pkg.develop(PackageSpec(path = pkgdir))
+        statusfile = joinpath(wrapperdir, "done.log")
+        open(joinpath(wrapperdir, "src", wrappername * ".jl"), "w") do io
+            println(
+                io,
+                """
 module $wrappername
 using $pkgname
 # Signal Aqua from the precompilation process that we've finished loading the package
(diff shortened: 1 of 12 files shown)


src/persistent_tasks.jl
```diff
@@ -0,0 +1,145 @@
...
+function test_persistent_tasks(package::Module; kwargs...)
+    test_persistent_tasks(PkgId(package); kwargs...)
+end
+
+"""
+    Aqua.test_persistent_tasks_deps(package; kwargs...)
+
+Test all the dependencies of `package` with [`Aqua.test_persistent_tasks`](@ref).
+"""
+function test_persistent_tasks_deps(package::PkgId; fails=Dict{String,Bool}(), kwargs...)
```
lgoettgens, line 90: The kwargs should be explained in the docstring (at least `fails` and that everything else is passed to `test_persistent_tasks`).

test/pkgs/PersistentTasks/PersistentTask/Project.toml
```diff
@@ -0,0 +1,4 @@
+name = "PersistentTask"
+uuid = "e5c298b6-d81d-47aa-a9ed-5ea983e22986"
+authors = ["Tim Holy <tim.holy@gmail.com>"]
+version = "0.1.0"
```
lgoettgens, line 4:
```suggestion
```
keep consistency with other pkgs in `test/`

test/pkgs/PersistentTasks/TransientTask/Project.toml
```diff
@@ -0,0 +1,4 @@
+name = "TransientTask"
+uuid = "94ae9332-58b0-4342-989c-0a7e44abcca9"
+authors = ["Tim Holy <tim.holy@gmail.com>"]
+version = "0.1.0"
```
lgoettgens, line 4:
```suggestion
```

test/pkgs/PersistentTasks/UsesBoth/Project.toml
```diff
@@ -0,0 +1,8 @@
+name = "UsesBoth"
+uuid = "96f12b6e-60f8-43dc-b131-049a88a2f499"
+authors = ["Tim Holy <tim.holy@gmail.com>"]
+version = "0.1.0"
```
lgoettgens, line 4:
```suggestion
```

test/test_persistent_tasks.jl
```diff
@@ -0,0 +1,28 @@
...
+    prj = TOML.parsefile(joinpath(path, "Project.toml"))
+    return PkgId(UUID(prj["uuid"]), prj["name"])
+end
+
+
+@testset "PersistentTasks" begin
+    Aqua.test_persistent_tasks(getid("TransientTask"))
+    Aqua.test_persistent_tasks_deps(getid("TransientTask"))
+
+    if all((Base.VERSION.major, Base.VERSION.minor) .>= (1, 10))
```
lgoettgens, line 21: Why not like this?
```suggestion
    if VERSION >= v"1.10"
```

lgoettgens (commented):
Overall, this looks like a good improvement for Aqua. Thanks! Unfortunately, the tests did not succeed and the reviewdog has some annotations. I have added some further remarks below.

fingolfin:
Thank @timholy . Unfortunately CI tests fail in Julia 1.4-1.10 due to a CI timeout -- it seems the tests get suck somehow

src/persistent_tasks.jl
```diff
@@ -0,0 +1,145 @@
...
+`Task`s in its `__init__` function may precompile successfully,
+but block precompilation of any packages that depend on it.
+
+# Example
+
+Let's create a dummy package, `PkgA`, that launches a persistent `Task`:
+
+```julia
+
+```julia
```
fingolfin, line 21:
```suggestion
```

src/persistent_tasks.jl
```diff
@@ -0,0 +1,145 @@
...
+to launch the tasks and cleanly shut them down.
+
+# Arguments
+- `package`: a top-level `Module` or `Base.PkgId`.
+
+# Keyword Arguments
+- `tmax::Real`: the maximum time (in seconds) to wait between loading the
+  package and forced shutdown of the precompilation process.
+"""
+function test_persistent_tasks(package::PkgId; tmax=5, fails::Bool=false)
```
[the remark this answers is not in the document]
timholy: It's 5s from the *end* of loading. I changed it to 10s, but given that the contents of the dummy package are trivial, writing the precompile cache files should be quite trivial.

src/persistent_tasks.jl
timholy: `fails` is just intended to support Aqua's internal tests, but sure.

timholy:
Now all passing except Windows. I will investigate later, other duties call.

src/persistent_tasks.jl
```diff
@@ -0,0 +1,156 @@
...
+    wrapperdir = tempname()
+    wrappername, _ = only(Pkg.generate(wrapperdir))
+    Pkg.activate(wrapperdir)
+    Pkg.develop(PackageSpec(path = pkgdir))
+    statusfile = joinpath(wrapperdir, "done.log")
+    open(joinpath(wrapperdir, "src", wrappername * ".jl"), "w") do io
+        println(
+            io,
+            """
+module $wrappername
```
JeffBezanson, line 125: This can be indented.

timholy: It was originally, but this package requires a JuliaFormatter test and it won't let me 🤣.

timholy:
Yay, green.

src/persistent_tasks.jl
```diff
@@ -0,0 +1,156 @@
...
+
+# Arguments
+- `package`: a top-level `Module` or `Base.PkgId`.
+
+# Keyword Arguments
+- `tmax::Real`: the maximum time (in seconds) to wait after loading the
+  package before forcibly shutting down the precompilation process (triggering
+  a test failure).
+"""
+function test_persistent_tasks(package::PkgId; tmax = 10, fails::Bool = false)
```
fingolfin, line 72: What is `fails` for? Just to simplify testing?

Other similar functions use a kwarg `broken` for this (and also document it), so to keep this consistent, it would be nice if that was the case here, too.

src/persistent_tasks.jl
```diff
@@ -0,0 +1,156 @@
...
+    @testset "$package persistent_tasks" begin
+        result = root_project_or_failed_lazytest(package)
+        result isa LazyTestResult && return result
+        @test fails ⊻ precompile_wrapper(result, tmax)
```
fingolfin, line 76: So perhaps like this (but obviously callers would still need to be adjusted)
```suggestion
- `broken::Bool = false`: If true, it uses `@test_broken` instead of
  `@test`.
"""
function test_persistent_tasks(package::PkgId; tmax = 10, broken::Bool = false)
    @testset "$package persistent_tasks" begin
        result = root_project_or_failed_lazytest(package)
        result isa LazyTestResult && return result
        if broken
            @test_broken precompile_wrapper(result, tmax)
        else
            @test precompile_wrapper(result, tmax)
        end
```

fingolfin (commented):
Thank you very much for this wonderful contribution.

I have one remaining concern: there is an established pattern for signalling a broken test case, using a `broken` kwarg; it would be nice if this PR also used it, instead of using `fails` for essentially the same purpose. Would you mind doing that? If not, I'll try to make the necessary adjustments but it may take me a while to find the time (sorry)

src/persistent_tasks.jl
timholy: I merged this, but then I changed it back to the `⊻`. The issue is that `@test_broken` usually marks something you want to fix in the package being tested, or a bug in Julia itself. In this case, it's more like `@test_throws` and we actually want it to fail---that's a sign that Aqua is doing its job correctly. So the xor logic seems correct.

timholy: Hmm, now I worry: was I right to use something different from `broken` at the outset? Is `broken` really intended for users of Aqua? `fails` was only there for testing, it's not intended to be used to mark tests in other packages as things you'd like to have pass, but don't.

timholy:
I changed the name to `broken` but I'm no longer sure that was the right thing to do. Is `broken` intended to used by packages that are using Aqua for their tests? I.e., "gosh, I want this Aqua test to pass, but it doesn't." In contrast, `fails` was purely for veracity of Aqua's own internal testing; it's not intended to be used by "consumers" of Aqua.

timholy:
I see a "needs changelog" label but I don't see a "NEWS" file?

lgoettgens:
> I see a "needs changelog" label but I don't see a "NEWS" file?

It has only been introduced this week. If you rebase your branch onto latest master, you should find a `CHANGELOG.md`.

lgoettgens:
I'll later have a look at your issue with `broken` and `fails`

lgoettgens:
I have slightly altered the interface. Many of the issues here can easily be solved by splitting test functions up into one computing something and another one checking it with `@test`. 
`has_persistent_tasks` now returns a boolean. `test_persistent_tasks` just checks this boolean (depending of a kwarg `broken` with `@test` or `@test_broken`). Internal testing just tests `has_persistent_tasks`.
Similar to `test_persistent_tasks_deps`: I changed it to a function returning a vector of the failing packages instead. This is almost as readable as an output but gives users access to the results (if they want to use it somehow), and makes testing this a lot cleaner.

Similar adaptations are happening one after the other to the other, already for longer time existing tasks as well. I hope these changes are fine for you. Please check that I didn't break everything. Thanks!

lgoettgens:
I noticed some caveats/errors with this new test. Aqua tests itself, and when enabling `test_persistent_tasks`, that task fails. That may possibly be some bug in the persisten_tasks code.

Furthermore, `find_persistent_tasks_deps` fails if a dependency is in the sysimg. For example, try to call it with `Aqua` as its argument.

src/persistent_tasks.jl
```diff
@@ -0,0 +1,160 @@
...
+    result = root_project_or_failed_lazytest(package)
+    result isa LazyTestResult && error("Unable to locate Project.toml")
+    return !precompile_wrapper(result, tmax)
+end
+
+"""
+    Aqua.find_persistent_tasks_deps(package; broken = Dict{String,Bool}(), kwargs...)
+
+Test all the dependencies of `package` with [`Aqua.test_persistent_tasks`](@ref).
+On Julia 1.10 and higher, you returns a list of all dependencies failing the test.
```
fingolfin, line 96: "you returns"?
```suggestion
On Julia 1.10 and higher, it returns a list of all dependencies failing the test.
```

lgoettgens:
Rebased this on top of the 0.7.3 release

lgoettgens (approved):
I found the culprits (see comments below). Furthermore I already pushed a quality of life change, that restores the active environment after the test.

@timholy please have a look, if this looks good to you. If the suggestions get committed, this is good to go from my pov.

timholy:
Sorry for my long absence here, and many thanks for carrying this forward. Just to check (since there have been many commits), are the comments in https://github.com/JuliaTesting/Aqua.jl/pull/174#pullrequestreview-1674378915 still relevant? They seem like nice changes to me. If you want I can update this, or feel free to do so yourself. Aside from some travel later this week, I am getting back to a point of having time to finish this off.

lgoettgens:
> Sorry for my long absence here, and many thanks for carrying this forward. Just to check (since there have been many commits), are the comments in [#174 (review)](https://github.com/JuliaTesting/Aqua.jl/pull/174#pullrequestreview-1674378915) still relevant? They seem like nice changes to me. If you want I can update this, or feel free to do so yourself. Aside from some travel later this week, I am getting back to a point of having time to finish this off.

Yes, these comments/suggestions are still relevant. I just rebase this onto master to resolve some conflicts. I think accepting all suggestions should resolve the remaining ci failures

lgoettgens:
You don't need to worry about the failures in julia 1.3 and before, as we will drop support anyway (see https://github.com/JuliaTesting/Aqua.jl/issues/216).

PR #183, opened 2023-09-18
Streamline uses of Compat.jl

lgoettgens:
The version numbers have been found using the approach in https://github.com/JuliaLang/Compat.jl/tree/release-3#developer-tips.


lgoettgens:
The docs failure may be fixed by #182. Let's wait for that and then try running it again.

fingolfin (approved):

PR #184, opened 2023-09-18
Fix TagBot setup

fingolfin:


lgoettgens:
Do you have access to the CI secrets @fingolfin? Or is this just guessing the name?
Afaik it is a common practice to use one ssh/deploy key for multiple things

lgoettgens:
Ah, this is probably due to the following:
> For example, if you use GitHub Actions to build the documentation for your package, you will find that with the default TagBot configuration, your documentation build is not triggered when a new tag is created. In this case, you will need to use [SSH Deploy Keys](https://github.com/JuliaRegistries/TagBot#ssh-deploy-keys).
Source: https://github.com/JuliaRegistries/TagBot#setup

fingolfin:
OK that would explain why this fix is needed.

And yes, I can see the names of the deploy key(s), so this is not just guess work.

PR #188, opened 2023-09-19
Fix `is_kwcall` for methods with type parameters

lgoettgens:
Fixes https://github.com/JuliaTesting/Aqua.jl/issues/186.

`is_kwcall` currently fails on julia 1.8 and older for methods where the third argument type is not a `DataType` but e.g. a `TypeVar`. This change only calls the erroring function in case that it is indeed a `DataType`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


lgoettgens:
If https://github.com/JuliaTesting/Aqua.jl/pull/189 gets merged before this, we need to add a changelog item.

fingolfin:
Changelog item needed (but I guess we would have needed one either way, the question is just who'd add it and when)

lgoettgens:
> Changelog item needed (but I guess we would have needed one either way, the question is just who'd add it and when)

Added it here. So this is ready to merge now

PR #191, opened 2023-09-21
Update the verion of Documenter.jl to v1.0

hyrodium:
This PR updates the version of Documenter.jl.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -8,7 +8,7 @@ makedocs(;
         repolink = "https://github.com/JuliaTesting/Aqua.jl",
     ),
     authors = "Takafumi Arakaki",
-    warnonly = true
+    warnonly = true,
 )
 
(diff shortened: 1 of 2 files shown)


docs/make.jl
```diff
@@ -4,7 +4,11 @@ makedocs(;
     modules = [Aqua],
     pages = ["Home" => "index.md"],
     sitename = "Aqua.jl",
+    format=Documenter.HTML(;
+        repolink = "https://github.com/JuliaTesting/Aqua.jl",
+    ),
     authors = "Takafumi Arakaki",
+    warnonly = true
```
hyrodium, line 11: This keyword argument `warnonly` is `false` by default, and it will raise errors if we have missing docstrings in the documentation. We are missing some docstrings in the document (e.g. `Aqua.undefined_exports`), so I set the value `false` to decrease the change in this PR.

fingolfin (approved):
Thanks

PR #197, opened 2023-09-25
Add more CI targets

lgoettgens:
Test on macos as well, as well as against julia nightly on both macos and windows to hopefully catch problems early.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,9 +30,17 @@ jobs:
           - 'nightly'
         include:
           - os: windows-latest
-            julia-version: 1
+            julia-version: '1'
           - os: windows-latest
             julia-version: '1.0'
+          - os: windows-latest
+            julia-version: 'nightly'
+          - os: macOS-latest
+            julia-version: '1'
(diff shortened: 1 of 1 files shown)


fingolfin:
targest -> targets

fingolfin (approved):
Of course this will make CI slower. Do you think it is likely that this pure Julia package will run into errors that only occur on macOS and/or Windows?

Anyway, other than waiting a bit longer for PRs to be ready, I see no drawbacks *shrug*

lgoettgens:
> Of course this will make CI slower. Do you think it is likely that this pure Julia package will run into errors that only occur on macOS and/or Windows?

There e.g. was an issue only occurring on Windows with the `persistent_tasks` PR which led to 43cbd3a93a213483bc8d4fceba1f21c4f697e99b. I would therefore like some few runners of all OSes.

> Anyway, other than waiting a bit longer for PRs to be ready, I see no drawbacks _shrug_

The JuliaTesting runners are usually not used, so that is basically negligible. (Apart from right now somehow)

Issue #178, opened 2023-09-12
Option for `test_deps_compat` to check extra/test deps have compat


nickrobinson251:
So that this test fails is the deps in `extras` don't have `compat`

e.g. this Project.toml would currently pass `test_deps_compat(FooBar)`
```julia
name = "FooBar"
uuid = "bb073782-7f75-46c6-bcf5-0b847ae15966"

[deps]
Qux = "62c31256-0984-4232-9d5d-7b366db4679b"

[compat]
Qux = "1"

[extras]
Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"

[targets]
test = ["Aqua", "Test"]
```
but it'd be great to have the option for this to fail with ```does not specify `compat` for: Aqua``` if we use e.g. `test_deps_compat(FooBar; check_extras=true)`

PR #200, opened 2023-09-27
Add option to `test_deps_compat` to test extras and weakdeps

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/178 and adds the same for `weakdeps` as well.

The tests are disabled by default, until maybe the next breaking release (see https://github.com/JuliaTesting/Aqua.jl/pull/202).

### DIFF
--- a/src/deps_compat.jl
+++ b/src/deps_compat.jl
@@ -9,12 +9,43 @@ each package listed under `deps`.
   them.
 
 # Keyword Arguments
+## Test choosers
+- `check_extras = false`: If true, additionally check "extras". A NamedTuple
+  can be used to pass keyword arguments with test options (see below).
+- `check_weakdeps = false`: If true, additionally check "weakdeps". A NamedTuple
+  can be used to pass keyword arguments with test options (see below).
+
+## Test options
+If these keyword arguments are set directly, they only apply to the standard test
+for "deps". To apply them to "extras" and "weakdeps", pass them as a NamedTuple
+to the corresponding `check_\$x` keyword argument.
 - `broken::Bool = false`: If true, it uses `@test_broken` instead of
-  `@test`.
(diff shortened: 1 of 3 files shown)


src/deps_compat.jl
```diff
@@ -9,12 +9,43 @@ each package listed under `deps`.
...
 
 # Keyword Arguments
+## Test choosers
+- `check_extras = false`: If true, additionally check "extras". A NamedTuple
+  can be used to pass keyword arguments with test options (see below).
+- `check_weakdeps = false`: If true, additionally check "weakdeps". A NamedTuple
+  can be used to pass keyword arguments with test options (see below).
+
+## Test options
+If these keyword arguments are set directly, they only apply to the standard test
```
gdalle, line 19: Can you think of a reason why you wouldn't want the same set of kwargs for all your deps checks?

gdalle:
This is a good idea!

test/test_smoke.jl
```diff
@@ -16,7 +16,7 @@ Aqua.test_all(
     undefined_exports = false,
     project_extras = false,
     stale_deps = false,
-    deps_compat = false,
+    deps_compat = (; check_extras = true, check_weakdeps = true),
```
gdalle, line 19: Those tests are empty for Aqua itself, which probably explains why the corresponding code lines are not hit

lgoettgens: The code does only detect that there are not extras and weakdeps way deeper in the code. And it's only the `@testset` lines that are not hit, everything inside is hit. So this is just an example of bad coverage detection

lgoettgens:
I'll now merge this although @fingolfin's approval is still missing (he said he currently doesn't have time to review it) to get forward with the 0.8 release.

Issue #136, opened 2023-06-05
Check weakdeps for being stale


lgoettgens:
> But we would also want to check for actual stale deps in the extensions too, right? So I guess the criterion is rather "is the dependency imported at all, whether in the main package or in an extension"

_Originally posted by @gdalle in https://github.com/JuliaTesting/Aqua.jl/issues/107#issuecomment-1572756617_

lgoettgens:
After thinking about this again, I think that this is a no-op:
For a package extension, say `PkgExt`, to be available its weakdeps have to be already loaded. But then there is nothing we can check for if they are indeed imported or used in `PkgExt`.

To recall the idea behind the usual `test_stale_deps`: Load the pkg, list all pkgs that are loaded after that, and report everything in the Project.toml that has not been loaded.

@gdalle Do you have any idea on how this could work conceptually? If not, I would close this.

gdalle:
Oh I didn't know it worked like that, in that case yeah it's tricky. Maybe add a warning to the Aqua docs saying that the deps of extensions cannot be checked?

PR #203, opened 2023-09-28
Add package extensions to `test_stale_deps` docstring

lgoettgens:
Closes https://github.com/JuliaTesting/Aqua.jl/issues/136 as proposed in https://github.com/JuliaTesting/Aqua.jl/issues/136#issuecomment-1738637286.

### DIFF
--- a/src/stale_deps.jl
+++ b/src/stale_deps.jl
@@ -5,7 +5,13 @@ Test that `package` loads all dependencies listed in `Project.toml`.
 Note that this does not imply that `package` loads the dependencies
 directly, this can be achieved via transitivity as well.
 
-!!! note "Known bug"
+!!! note "Weak dependencies and extensions"
+
+    Due to the automatic loading of package extensions once all of
+    their dependencies are loaded, a package extension can, by design,
+    not have any stale dependencies.
+
+!!! warning "Known bug"
 
     Currently, `Aqua.test_stale_deps` does not detect stale
     dependencies when they are stdlib.  This is considered a bug and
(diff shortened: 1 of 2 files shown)


lgoettgens:
@gdalle Is this what you had in mind?

src/stale_deps.jl
```diff
@@ -5,7 +5,13 @@ Test that `package` loads all dependencies listed in `Project.toml`.
 Note that this does not imply that `package` loads the dependencies
 directly, this can be achieved via transitivity as well.
 
-!!! note "Known bug"
+!!! note "Weak dependencies and extensions"
+
+    Due to the automatic loading of package extensions once all of
```
gdalle, line 10: For me it's not that they can't have any stale dependencies, it's that we cannot check for them with Aqua.

As a side note, the dependencies of an extension (the `using` statements at the top) are a superset of its triggers (that are necessary for it to load) because they also include the dependencies of the parent package itself

src/stale_deps.jl
```diff
@@ -5,7 +5,13 @@ Test that `package` loads all dependencies listed in `Project.toml`.
...
+    their dependencies are loaded, a package extension can, by design,
+    not have any stale dependencies.
```
lgoettgens, line 12:
```suggestion
    their trigger dependencies are loaded, Aqua.jl can, by design of julia,
    not check if a package extension indeed loads all of its trigger dependencies using `import` or `using`. 
```

@gdalle I understand your point now better. What do you think about this?

gdalle: yeah I think this is better!

Issue #79, opened 2022-06-16
How to skip ambiguities of keyword functions?


ranocha:
Thanks for this great package! When trying to set up Aqua.jl tests for one packages, I ran into
```
Ambiguity #2
(::Base.var"#isapprox##kw")(::Any, ::typeof(isapprox), p1::Polynomials.AbstractPolynomial{T}, n::S) where {T, S} in Polynomials at /afs/math.uni-hamburg.de/users/am/bbb3094/.julia/packages/Polynomials/HJBX9/src/common.jl:1077
(::Base.var"#isapprox##kw")(::Any, ::typeof(isapprox), ::Any, ::Missing) in Base at missing.jl:91

Possible fix, define
  (::Base.var"#isapprox##kw")(::Any, ::typeof(isapprox), ::Polynomials.AbstractPolynomial{T}, ::Missing) where T

Ambiguity #3
(::Base.var"#isapprox##kw")(::Any, ::typeof(isapprox), n::S, p1::Polynomials.AbstractPolynomial{T}) where {T, S} in Polynomials at /afs/math.uni-hamburg.de/users/am/bbb3094/.julia/packages/Polynomials/HJBX9/src/common.jl:1087
(::Base.var"#isapprox##kw")(::Any, ::typeof(isapprox), ::Missing, ::Any) in Base at missing.jl:90

Possible fix, define
  (::Base.var"#isapprox##kw")(::Any, ::typeof(isapprox), ::Missing, ::Polynomials.AbstractPolynomial{T}) where T
```
How can I tell Aqua.jl to ignore method ambiguities of this keyword function? Passing `exclude=[isapprox]` or `exclude=[isapprox, Base.var"#isapprox##kw"]` does not help.

lgoettgens:
Could you give some more information, like which package this is about and on which julia version the above output was produced?

In general, you can get the kwsorter of a function `foo` by `Core.kwfunc(foo)`. However, since julia 1.9.0-DEV.1598, the individual kwsorters have been removed in favor of one function handling all cases. So it seems that, unfortunately, you can in julia post-1.9 only exclude all `kwcall` ambiguities or none.

ranocha:
It's BSeries.jl, see 
https://github.com/ranocha/BSeries.jl/blob/70da9b93fca0be38494b4a76ca37f62a33bbb460/test/runtests.jl#L2162-L2189

I don't remember the Julia version I used more than a year ago - probably the latest stable Julia release at that time

[...]

PR #204, opened 2023-09-28
Enhance `test_ambiguities`' exclude options for functions with kwargs

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/79.

After this change, excluding some `foo` additionally excludes all instances of `Core.kwfunc(foo)` that belong to `foo` (i.e. the third parameter of the signature is `typeof(foo)`.


fingolfin (approved):
Looks good to me

PR #209, opened 2023-10-13
Remove test_project_toml_formatting

### LINKED ISSUE #208: Project.toml formatting test not sensible in julia pre-1.7
`Pkg.jl` my alter a `Project.toml` file when instantiating before julia 1.7 (cf. https://github.com/JuliaLang/Pkg.jl/issues/3481#issuecomment-1637811994).
Thus, the current test is basically a no-op returning `pass` on julia 1.6 and before (in CI), as the `Project.toml` gets formatted when instantiating the project and thus everything is already formatted when Aqua inspects it.

If a package uses package extensions, this is even worse, as the automatic formatting on julia 1.6 and before is incompatible with the one required by the Aqua test. Therefore, this test always fails in these cases.

For the previous discussion, please refer to https://github.com/JuliaTesting/Aqua.jl/issues/105#issuecomment-1551065724 and the comments below it.

fingolfin:
It stands out from all other tests in Aqua in that it does not actually deal with 'quality'. The benefit is tiny if there is one at all, at the same time it is unreasonably compicated to implement and use right.

Resolves #208
Closes #206, closes #210


lgoettgens (approved):

lgoettgens:
@fingolfin I found some more code that is no longer needed. Could you have a look?

fingolfin:
@lgoettgens thank you, makes sense, and Looks good to me

PR #210, opened 2023-10-13
Restrict `project_toml_formatting` to julia 1.7 and later

### LINKED ISSUE #208: Project.toml formatting test not sensible in julia pre-1.7

lgoettgens:
Resolves #208
Closes #206, closes #209.

Just make the `test_toml_formatting` a no-op on julia 1.6 and before.

### DIFF
--- a/src/project_toml_formatting.jl
+++ b/src/project_toml_formatting.jl
@@ -1,10 +1,17 @@
 """
     Aqua.test_project_toml_formatting(packages)
+
+Test that the files `Project.toml`, `docs/Project.toml` (if present),
+and `test/Project.toml` (if present) are in canonical format according to Pkg.jl v1.9.
+
+This test does nothing on julia v1.6 or earlier.
 """
 function test_project_toml_formatting(packages)
-    @testset "$(result.label)" for result in analyze_project_toml_formatting(packages)
-        @debug result.label result
-        @test result ⊜ true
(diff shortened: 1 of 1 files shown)


gdalle (approved):

gdalle (changes requested):
Sorry I didn't mean to approve right away: the code changes are good but could you add tests, like the ones i tried to whip up in #206?

lgoettgens:
> Sorry I didn't mean to approve right away: the code changes are good but could you add tests, like the ones i tried to whip up in #206?

I would do that once the decision between this and #210 is made. I'll try to talk to @fingolfin this week about it.

PR #211, opened 2023-10-13
Add CI job checking for bad PR labels

lgoettgens:
Basically a copy of https://github.com/JuliaLang/julia/pull/48359. I think this is worth to try.


lgoettgens:
@fingolfin can you mark this new job as "required"?

PR #213, opened 2023-10-16
Better GitHub Action workflow for JuliaFormatter

hyrodium:
x-ref: https://github.com/JuliaTesting/Aqua.jl/pull/128#issuecomment-1764625488

I have just copied the YAML file from https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml.

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -1,10 +1,27 @@
-name: Code style
+name: Format suggestions
 
 on:
   pull_request:
 
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
 jobs:
-  code-style:
+  format:
     runs-on: ubuntu-latest
     steps:
-      - uses: tkf/julia-code-style-suggesters@v1
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1
+      - run: |
+          julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
+          julia  -e 'using JuliaFormatter; format("."; verbose=true)'
+      - uses: reviewdog/action-suggester@v1
+        with:
+          tool_name: JuliaFormatter
+          fail_on_error: true
+          filter_mode: added


hyrodium:
The original workflow [julia-code-style-suggesters](https://github.com/tkf/julia-code-style-suggesters) is a combination of:

* JuliaFormatter.jl
* JuliaProjectFormatter.jl
* reviewdog

All features in [JuliaProjectFormatter.jl](https://github.com/tkf/JuliaProjectFormatter.jl) seems a subset of Aqua.jl, so there would be no problem in this PR. (https://github.com/tkf/JuliaProjectFormatter.jl/pull/7#issuecomment-1763269610)

hyrodium:
[The log](https://github.com/JuliaTesting/Aqua.jl/actions/runs/6560983685/job/17819735269?pr=218) shows the folloing message, so it seems we need more permissions for that token.

>reviewdog: This GitHub token doesn't have write permission of Review API [1], 
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
[1]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target, 
[2]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#logging-commands

lgoettgens:
I applied @fingolfin's suggestions from #218 here.

lgoettgens:
`reviewdog: environment variable $REVIEWDOG_GITHUB_API_TOKEN is not set` https://github.com/JuliaTesting/Aqua.jl/actions/runs/6627926585/job/18003929859?pr=213#step:5:185

@fingolfin something seems to be wrong

hyrodium:
I have applied the commits from #218. (x-ref https://github.com/JuliaTesting/Aqua.jl/pull/218#issuecomment-1779121854)

hyrodium:
We still have the following message:

>reviewdog: This GitHub token doesn't have write permission of Review API [1], 

https://github.com/JuliaTesting/Aqua.jl/actions/runs/6642425658/job/18047160049?pr=213#step:5:192

hyrodium:
We can use `julia-actions/julia-format@v2` when https://github.com/julia-actions/julia-format/pull/24 is merged.

lgoettgens:
With the `julia-format@v2` action, there is still the same error as in https://github.com/JuliaTesting/Aqua.jl/pull/213#issuecomment-1773660514.

hyrodium:
Hmm, the suggestion works file in other PRs (https://github.com/JuliaTesting/Aqua.jl/pull/255#discussion_r1408579106) now, so I'm not sure the reason for that error.
Perhaps, recreating this PR would solve the problem?

lgoettgens:
> Hmm, the suggestion works file in other PRs ([#255 (comment)](https://github.com/JuliaTesting/Aqua.jl/pull/255#discussion_r1408579106)) now, so I'm not sure the reason for that error. Perhaps, recreating this PR would solve the problem?

The only difference to other PRs that I can see is that this PR here is from a fork.

hyrodium:
>The only difference to other PRs that I can see is that this PR here is from a fork.

That's correct. I found an issue about this: https://github.com/reviewdog/action-suggester/issues/19

Issue #212, opened 2023-10-15
Adapt to change in compat requirements in the general registry


lgoettgens:
See https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958
Aqua.jl should change the tests accordingly.

lgoettgens:
I'll try to get this finished the upcoming week, so that we can release v0.8.0 shortly after.

PR #215, opened 2023-10-17
Remove special handling of stdlibs in `test_deps_compat`

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/212.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The minimum requirement for julia was raised from `1.0` to `1.4`. ([#221](https://github.com/JuliaTesting/Aqua.jl/pull/221))
 - In `test_deps_compat`, the two subtests `check_extras` and `check_weakdeps` are now run by default. ([#202](https://github.com/JuliaTesting/Aqua.jl/pull/202)) [BREAKING]
+- `test_deps_compat` now reqiures compat entries for all dependencies. Stdlibs no longer get ignored. This change is motivated by similar changes in the General registry. ([#215](https://github.com/JuliaTesting/Aqua.jl/pull/215)) [BREAKING]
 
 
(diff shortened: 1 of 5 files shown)


fingolfin (approved):

vtjnash:
This (or something surrounding it) seems to have caused quite a few failures in PkgEval, which reduces the overall testing coverage in the ecosystem. Is there something we can do about this? For example:
https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/1fc9d54_vs_72cd63c/Quadmath.primary.log
https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/1fc9d54_vs_72cd63c/AbstractTrees.primary.log
https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/1fc9d54_vs_72cd63c/ManualMemory.primary.log

Looks at the apparent red cliff in the report: https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/1fc9d54_vs_72cd63c/report.html

Some don't seem precisely related, as the error appears to complain about packages (e.g. BenchmarkTools) that don't appear to even be installed in the log?

lgoettgens:
Yes, parts of the failures you linked are due to this change. But a lot are due to checking extras and weakdeps in general since #202. So every package declaring Aqua as a extra-Dep without a compat bound will fail as well.

The clean solution would be to change all failing packages to either fix the issues there or (easier) add a compat entry for Aqua 0.7 there.
If you are interested in PkgEval in particular, I can think about changing some kwargs in the default test list depending on the PkgEval ENV variable, so by default PkgEval testing would only use a subset of Aqua tests. Would that be something you are interested in? If the list of failing packages is rather short, I think this is not worth the effort and we should instead provide PRs to them. I unfortunately don't really feel confident to condense such a list from the PkgEval report you linked above.

Issue #216, opened 2023-10-17
Drop support for julia 1.0-1.3?


lgoettgens:
https://github.com/JuliaTesting/Aqua.jl/pull/214 and thus https://github.com/JuliaTesting/Aqua.jl/pull/215 needed a workaround to circumvent an old problem in Pkg.jl in connection with Test.jl.
This would furthermore enable us to merge https://github.com/JuliaTesting/Aqua.jl/pull/195 eventually, thus removing a large code duplication with julia base.

> or we just drop support for older Julia versions in Aqua.jl ? I am not sure how valuable it is? After all, as was pointed out, people on older Aqua versions can still use the old Aqua releases. If they want any new features, well, they have to upgrade Julia.
There may be some people who absolutely must stick on Julia 1.0, but that's seems like a rather fringe demographic?
> -by @fingolfin on slack

I, personally, am in favor of this change. If somebody has some good reason why we should not proceed, please let me know.

hyrodium:
I think we can also drop versions before v1.6 if we drop support for v1.0 because the other versions such as v1.4 are not LTS.

fingolfin:
Also fine by me.

As it is, I wonder if it's worth bending over backwards if we don't even know whether anyone uses it...

One could of course also do something like this: the next release  prints a notice when loaded in Julia <= 1.3 (or 1.5) saying something like "if you'd like to receive updates for Aqua.jl in Julia $(VERSION) then please let us know at ISSUEURL". Then we wait a bit and see if someone reacts.

lgoettgens:
> One could of course also do something like this: the next release prints a notice when loaded in Julia <= 1.3 (or 1.5) saying something like "if you'd like to receive updates for Aqua.jl in Julia $(VERSION) then please let us know at ISSUEURL". Then we wait a bit and see if someone reacts.

I don't know if many users look at ci logs when they succeed. And making the go red would be a breaking change. So no idea how to really do that

PR #224, opened 2023-10-27
Revert "Drop support for julia 1.0-1.3 (#221)"

lgoettgens:


lgoettgens:
As already expected, `Test = "1.0" fails on julia 1.0-1.3 ([log](https://github.com/JuliaTesting/Aqua.jl/actions/runs/6664533436?pr=224))

lgoettgens:
`Test = "<0.0.1, 1"` indeed seems to work. Thanks @DilumAluthge!

lgoettgens:
@fingolfin What do you think? Should we enable compatibility with julia 1.0 again (by merging this PR)? If yes, I should look through the changes in detail, because there were a few conflicts when reverting stuff.

If you approve this, could you change the required CI jobs back to needing 1.0-1.3 for ubuntu, 1.0 for windows and macOS, and instead remove 1.4 for windows and macOS?

hyrodium:
I believe it is important that the code written for Julia 1.0 works with the latest versions of Julia and some packages. On the other hand, I think it is rare for users who do not upgrade their Julia version to only upgrade the version of Aqua.jl, so I feel there is no need to actively support Julia versions prior to 1.6 (LTS).

Sorry for repeating my previous comment (https://github.com/JuliaTesting/Aqua.jl/issues/216#issuecomment-1766572091), but I feel it would be nice to reduce the costs to maintain this package.

fingolfin (approved):
Honestly I don't think support for the older Julia versions is worth the hassle.

However, if you are willing to take care of the burden, I certainly won't object! In the worst case, we could still re-remove it later.

As such, I think you should make the final decision, Lars. I won't complain either way :-)

src/persistent_tasks.jl
```diff
@@ -61,6 +61,8 @@ end
 In more complex cases, you may need to set up independently-callable functions
 to launch the tasks and set conditions that allow them to cleanly exit.
 
+On julia version 1.9 and before, this test always succeeds.
```
fingolfin, line 64:
```suggestion
On Julia version 1.9 and before, this test always succeeds.
```

PR #233, opened 2023-11-03
Sort ambiguities and type piracies by function name

### LINKED ISSUE #226: Feature request: group ambiguities by function
Right now ambiguities seem to be reported in a seemingly arbitrary order (perhaps the order they are encountered in?)

I am looking at a project which has [130 ambiguities](https://github.com/Nemocas/Nemo.jl/issues/1553) (too many *sigh*). To get a better overview, it would be kinda nice if the ambiguities were group by function: i.e. first all ambiguities in function `aaaaa` (if that existed), then those in `bbbbb`, etc. (just using `sort` on the function names -- if they are "weird" or involve `kwarg` or whatever, fine, sort them whatever way, but however it is it won't be worse than what we have right now, IMHO).

lgoettgens:
Resolves #226.

Sorts both the reported ambiguities and type piracies by function name for easier fixing.


fingolfin (approved):
Great!

PR #234, opened 2023-11-04
Improve error when [extras] and test/Project.toml inconsistent

fingolfin:
It is very easy to miss that this error only occurs because of a
missing compat entry for Julia...


lgoettgens:
Could you please make the formatter job happy and add a one-liners to the changelog? Then this is good to go from my POV

Issue #235, opened 2023-11-04
Require Julia compat bound


cossio:
Currently Aqua does not fail if there is no Julia compat bound in `Project.toml`. Instead it just assumes that the project intentionally supports older Julia versions (e.g., see https://github.com/JuliaTesting/Aqua.jl/pull/234).

Would it be better to enforce an explicit Julia compat bound?

PR #236, opened 2023-11-06
Require `[compat]` entry for `julia`

lgoettgens:
Resolves https://github.com/JuliaTesting/Aqua.jl/issues/235.

As this is point 7 in the [guidelines for the General registry](https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/#New-packages), I support Aqua checking this.

### DIFF
--- a/src/deps_compat.jl
+++ b/src/deps_compat.jl
@@ -2,7 +2,7 @@
     Aqua.test_deps_compat(package)
 
 Test that the `Project.toml` of `package` has a `compat` entry for
-each package listed under `deps`.
+each package listed under `deps` and for `julia`.
 
 # Arguments
 - `packages`: a top-level `Module`, a `Base.PkgId`, or a collection of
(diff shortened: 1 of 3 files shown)


lgoettgens:
@fingolfin Do you wanna have a look at this? If not, I will just merge this and https://github.com/JuliaTesting/Aqua.jl/pull/223, and then release 0.8.0.

src/deps_compat.jl
```diff
@@ -23,18 +24,29 @@ to the corresponding `check_\$x` keyword argument.
...
 """
-function test_deps_compat(pkg::PkgId; check_extras = true, check_weakdeps = true, kwargs...)
+function test_deps_compat(
+    pkg::PkgId;
+    check_julia = true,
+    check_extras = true,
+    check_weakdeps = true,
+    kwargs...,
+)
+    if check_julia !== false
```
hyrodium, line 34:
```suggestion
    check_julia::Bool = true,
    check_extras::Bool = true,
    check_weakdeps::Bool = true,
    kwargs...,
)
    if check_julia
```

I think argument-type specification is better for readability and error handling.

hyrodium (commented):
Added a minor comment. Other changes LGTM.

src/deps_compat.jl
lgoettgens: That is not possible here in an easy way, as one can pass a `Bool` or a `NamedTuple`. The docstring says `A NamedTuple can be used to pass keyword arguments with test options (see below).`

hyrodium (approved):

PR #239, opened 2023-11-15
Add `find_persistent_task_deps` to doc build

timholy:
Also adds a cross-ref from `test_persistent_tasks`.

CC @NHDaly


lgoettgens (approved):
Looks good. Thanks!

PR #250, opened 2023-11-20
Overhaul documentation

lgoettgens:
- [x] Split into one doc page per test

Structurize, fill with content etc.:
- [x] ambiguities
- [x] deps_compat
- [x] exports
- [x] persistent_tasks
- [x] piracies
- [x] project_extras
- [x] stale_deps
- [x] unbound_args
- [x] test_all

Preview at https://juliatesting.github.io/Aqua.jl/previews/PR250/

### DIFF
--- a/docs/src/ambiguities.md
+++ b/docs/src/ambiguities.md
@@ -2,8 +2,9 @@
 
 Method ambiguities are cases where multiple methods are applicable to a given set of arguments, without having a most specific method.
 
+## Examples
 One easy example is the following:
-```julia
+```@example
 f(x::Int, y::Integer) = 1
 f(x::Integer, y::Int) = 2
 
--- a/docs/src/ambiguities.md
+++ b/docs/src/ambiguities.md
@@ -1,5 +1,21 @@
 # Ambiguities
 
+Method ambiguities are cases where multiple methods are applicable to a given set of arguments, without having a most specific method.
+
+One easy example is the following:
+```julia
+f(x::Int, y::Integer) = 1
+f(x::Integer, y::Int) = 2
+
+println(f(1, 2))
+```
+This will throw an `MethodError` because both methods are equally specific. The solution is to add a third method:
+```julia
+f(x::Int, y::Int) = ? # `?` is dependent on the use case, most times it will be `1` or `2`
+```
+
+## [Test function](@id test_ambiguities)
+
 ```@docs
 Aqua.test_ambiguities
 ```
--- a/docs/src/ambiguities.md
+++ b/docs/src/ambiguities.md
@@ -1,0 +1,5 @@
+# Ambiguities
+
+```@docs
+Aqua.test_ambiguities
+```
--- a/docs/src/ambiguities.md
+++ b/docs/src/ambiguities.md
@@ -4,7 +4,7 @@ Method ambiguities are cases where multiple methods are applicable to a given se
 
 ## Examples
 One easy example is the following:
-```@example
+```@repl
 f(x::Int, y::Integer) = 1
 f(x::Integer, y::Int) = 2
 
--- a/docs/src/unbound_args.md
+++ b/docs/src/unbound_args.md
@@ -27,6 +27,8 @@ h() = do_something(T[])
 h(x1::T, x2::T...) = do_something(T[x1, x2...])
 ```
 
+## [Test function](@id test_unbound_args)
+
 ```@docs
 Aqua.test_unbound_args
 ```
--- a/docs/src/unbound_args.md
+++ b/docs/src/unbound_args.md
@@ -1,0 +1,5 @@
+# Unbound Type Parameters
+
+```@docs
+Aqua.test_unbound_args
+```
--- a/docs/src/unbound_args.md
+++ b/docs/src/unbound_args.md
@@ -1,5 +1,32 @@
 # Unbound Type Parameters
 
+An unbound type parameter is a type parameter with a `where`,
+that does not occur in the signature of some dispatch of the method.
+
+## Examples
+
+The following methods each have `T` as an unbound type parameter:
+
+```julia
+f(x::Int) where {T} = do_something(x)
+
+g(x::S) where {S < : Number, T <: Number} = do_something(x)
+
+h(x::T...) where {T} = do_something(x)
+```
+
+In the cases of `f` and `g` above, the unbound type parameter `T` is neither
+present in the signature of the methods nor as a bound of another type parameter.
+Here, the type parameter `T` can be removed without changing any semantics.
+
+For signatures with `Vararg` (cf. `h` above), the type parameter is unbound for the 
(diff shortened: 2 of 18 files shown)


docs/src/ambiguities.md
```diff
@@ -0,0 +1,21 @@
+# Ambiguities
+
+Method ambiguities are cases where multiple methods are applicable to a given set of arguments, without having a most specific method.
+
```
hyrodium, line 4:
```suggestion

## Examples
```
Adding example section would be better for consistency with `unbound_args.md` and `piracies.md` pages.

docs/src/ambiguities.md
```diff
@@ -0,0 +1,21 @@
...
+One easy example is the following:
+```julia
+f(x::Int, y::Integer) = 1
+f(x::Integer, y::Int) = 2
+
+println(f(1, 2))
+```
```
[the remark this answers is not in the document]
lgoettgens:
````suggestion
```@example
f(x::Int, y::Integer) = 1
f(x::Integer, y::Int) = 2

println(f(1, 2))
```
````

File: docs/src/ambiguities.md [hunk 3]
```diff
@@ -0,0 +1,22 @@
+# Ambiguities
+
+Method ambiguities are cases where multiple methods are applicable to a given set of arguments, without having a most specific method.
+
+## Examples
+One easy example is the following:
+```@example
```
lgoettgens, line 7:
```suggestion
```@repl
```

docs/src/unbound_args.md
```diff
@@ -0,0 +1,34 @@
...
+that does not occur in the signature of some dispatch of the method.
+
+## Examples
+
+The following methods each have `T` as an unbound type parameter:
+
+```julia
+f(x::Int) where {T} = do_something(x)
+
+g(x::S) where {S < : Number, T <: Number} = do_something(x)
```
hyrodium, line 13: Fix syntax error
```suggestion
g(x::S) where {S <: Number, T <: Number} = do_something(x)
```

docs/src/unbound_args.md
```diff
@@ -0,0 +1,34 @@
...
+In the cases of `f` and `g` above, the unbound type parameter `T` is neither
+present in the signature of the methods nor as a bound of another type parameter.
+Here, the type parameter `T` can be removed without changing any semantics.
+
+For signatures with `Vararg` (cf. `h` above), the type parameter is unbound for the 
+zero-argument case (e.g. `h()`). A possible fix would be to replace `h` by two
+methods
+```julia
+h() = do_something(T[])
+h(x1::T, x2::T...) = do_something(T[x1, x2...])
```
hyrodium, line 27: We need `where {T}` here.
```suggestion
h(x1::T, x2::T...) where {T} = do_something(T[x1, x2...])
```

docs/src/unbound_args.md
```diff
@@ -0,0 +1,34 @@
...
+```
```
hyrodium, line 28: I think the example can be rewritten like this.

* Example functions `f` and `g` are really similar, so `g` can be removed.
* Use `@repl` block.
* Use `println(T)` instead of `do_something(x)` in `g` for tangible code.
* Add an error example with `ERROR: UndefVarError: `T` not defined`.

````suggestion
```@repl
f(x::Int) where {T} = do_something(x)
g(x::T...) where {T} = println(T)
```

In the cases of `f` above, the unbound type parameter `T` is neither
present in the signature of the methods nor as a bound of another type parameter.
Here, the type parameter `T` can be removed without changing any semantics.

For signatures with `Vararg` (cf. `g` above), the type parameter is unbound for the 
zero-argument case (e.g. `g()`).

```@repl
g(1.0, 2.0)
g(1)
g()
```

A possible fix would be to replace `g` by two methods.

```@repl
h() = println(Int)  # Defaults to `Int`
h(x1::T, x2::T...) where {T}= println(T)
h(1.0, 2.0)
h(1)
h()
```
````

Comment (lgoettgens) [reply docs/src/unbound_args.md:28]:
````suggestion
```@repl
f(x::Int) where {T} = do_something(x)
g(x::T...) where {T} = println(T)
```

In the cases of `f` above, the unbound type parameter `T` is neither
present in the signature of the methods nor as a bound of another type parameter.
Here, the type parameter `T` can be removed without changing any semantics.

For signatures with `Vararg` (cf. `g` above), the type parameter is unbound for the 
zero-argument case (e.g. `g()`).

```@repl
g(1.0, 2.0)
g(1)
g()
```

A possible fix would be to replace `g` by two methods.

```@repl
g() = println(Int)  # Defaults to `Int`
g(x1::T, x2::T...) where {T} = println(T)
g(1.0, 2.0)
g(1)
g()
```
````

File: docs/src/unbound_args.md (hunk 4 above)
Comment (lgoettgens) [reply docs/src/unbound_args.md:13]: superseeded by comment below

File: docs/src/unbound_args.md (hunk 5 above)
Comment (lgoettgens) [reply docs/src/unbound_args.md:27]: superseded by comment below

File: docs/src/unbound_args.md [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+
+```@repl
+g(1.0, 2.0)
+g(1)
+g()
+```
+
+A possible fix would be to replace `g` by two methods.
+
+```@repl
```
hyrodium, line 30: Currently, the documentation shows `g not defined`.
image

````suggestion
```@repl unbound
f(x::Int) where {T} = do_something(x)
g(x::T...) where {T} = println(T)
```

In the cases of `f` above, the unbound type parameter `T` is neither
present in the signature of the methods nor as a bound of another type parameter.
Here, the type parameter `T` can be removed without changing any semantics.

For signatures with `Vararg` (cf. `g` above), the type parameter is unbound for the 
zero-argument case (e.g. `g()`).

```@repl unbound
g(1.0, 2.0)
g(1)
g()
```

A possible fix would be to replace `g` by two methods.

```@repl unbound
````

After the fix, the function `g` has three methods, and the first method will never be called.
Is this okay?
This is why I suggest using `h` in [the suggestion](https://github.com/JuliaTesting/Aqua.jl/pull/250#discussion_r1408706061).

Comment (lgoettgens) [reply docs/src/unbound_args.md:30]:
````suggestion
```@repl unbound
f(x::Int) where {T} = do_something(x)
g(x::T...) where {T} = println(T)
```

In the cases of `f` above, the unbound type parameter `T` is neither
present in the signature of the methods nor as a bound of another type parameter.
Here, the type parameter `T` can be removed without changing any semantics.

For signatures with `Vararg` (cf. `g` above), the type parameter is unbound for the 
zero-argument case (e.g. `g()`).

```@repl unbound
g(1.0, 2.0)
g(1)
g()
```

A possible fix would be to replace `g` by two methods.

```@repl unbound2
````

That's true. I would like to keep the name the same though (as it states to `replace`). So I would make a different repl env instead.

hyrodium (approved):
LGTM!

PR #255, opened 2023-11-29
test_persistent_tasks: Add an optional `expr` to run in the precompile package

NHDaly:


NHDaly:
Okay! Thanks @lgoettgens.

I rebased on master, fixed the merge conflicts, added a test case, and changed `expr` to a keyword argument.

The reason I had made it a positional argument at first was because I wasn't sure what a good name should be. Is `expr` okay? Should it be `precompile_expr`?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.8.4] - 2023-11-30
+
+### Added
+
+- `test_persistent_tasks` now accepts an optional `expr` to run in the precompile package
```
lgoettgens, line 14:
```suggestion
- `test_persistent_tasks` now accepts an optional `expr` to run in the precompile package. ([#255](https://github.com/JuliaTesting/Aqua.jl/pull/255))
```
add link to PR

docs/src/persistent_tasks.md
```diff
@@ -32,6 +32,28 @@ fails to precompile: `using PkgA` runs `PkgA.__init__()`, which
...
+    # Code to run after loading MyPackage
+    server = MyPackage.start_server()
+    MyPackage.stop_server!(server)
+end)
+```
+
+here is an example test with a dummy expr which will obviously fail, because it's explicitly
+spawning a Task that never dies!
+```julia
+# This will fail because of the dangling spawn.
```
lgoettgens, line 51:
```suggestion
Here is an example test with a dummy expr which will obviously fail, because it's explicitly
spawning a Task that never dies.
```@repl
```
some reformulation, adding the error message of the failure

src/persistent_tasks.jl
```diff
@@ -16,6 +20,7 @@ On Julia version 1.9 and before, this test always succeeds.
 - `tmax::Real = 5`: the maximum time (in seconds) to wait after loading the
   package before forcibly shutting down the precompilation process (triggering
   a test failure).
+- `expr = nothing`: An expression to run in the precompile package.
```
lgoettgens, line 23:
```suggestion
- `expr::Expr = quote nothing end`: An expression to run in the precompile package.
```
update docstring accordingly

src/persistent_tasks.jl
```diff
@@ -29,10 +34,10 @@ function test_persistent_tasks(package::Module; kwargs...)
     test_persistent_tasks(PkgId(package); kwargs...)
 end
 
-function has_persistent_tasks(package::PkgId; tmax = 10)
+function has_persistent_tasks(package::PkgId; expr = nothing, tmax = 10)
```
lgoettgens, line 37:
```suggestion
function has_persistent_tasks(package::PkgId; expr::Expr = quote nothing end, tmax = 10)
```
Add type annotation

lgoettgens (approved):
Thanks for the changes! I have some minor comments, could you have a look at them? Once they are resolved, I'll update this PR for a version bump and do the merge/release procedure.

lgoettgens:
> The reason I had made it a positional argument at first was because I wasn't sure what a good name should be. Is `expr` okay? Should it be `precompile_expr`?

I have no real preference between the two.

src/persistent_tasks.jl
NHDaly: thank you!

src/persistent_tasks.jl
```diff
@@ -29,10 +34,10 @@ function test_persistent_tasks(package::Module; kwargs...)
     test_persistent_tasks(PkgId(package); kwargs...)
 end
 
-function has_persistent_tasks(package::PkgId; tmax = 10)
+function has_persistent_tasks(package::PkgId; expr::Expr = quote nothing end, tmax = 10)
```
NHDaly, line 37: Actually, I think this is even simpler:
```suggestion
function has_persistent_tasks(package::PkgId; expr::Expr = :(), tmax = 10)
```

NHDaly (commented):
Okay I think this is all set! Thanks for the nice review 😊 ❤️

NHDaly:
Thanks! Should be all done then

PR #256, opened 2023-11-29
persistent_tasks: suppress logs from Pkg.jl

aviatesk:
By using `io=devnull` option, which is only available in Julia 1.9 and above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.8.3] - 2023-11-29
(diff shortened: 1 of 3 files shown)


lgoettgens:
Thanks for your PR!
The whole test is only run for julia 1.10 and later (see https://github.com/JuliaTesting/Aqua.jl/pull/256/files#diff-f727412b094adb6463b973046def5a3004e1e292aca989b8b4564ce26f5132feR121), so you don't need all of these `@static if` everywhere.

aviatesk:
Thanks for your review! Updated.

lgoettgens (approved):
Looks great now, thanks! I'll proceed with this and a changelog entry.

PR #266, opened 2024-03-05
Update stale deps stdlibs known bug warning

LilithHafner:
I just got this error

```
Stale dependencies: Test Failed at /home/x/.julia/packages/Aqua/9p8ck/src/stale_deps.jl:31
  Expression: isempty(stale_deps)
   Evaluated: isempty(Base.PkgId[Base.PkgId(Base.UUID("9e88b42a-f829-5b0c-bbe9-9e923198166b"), "Serialization")])
```

So I think this bug is fixed.

Would tests be expected for this PR?

### DIFF
--- a/src/stale_deps.jl
+++ b/src/stale_deps.jl
@@ -12,13 +12,6 @@ directly, this can be achieved via transitivity as well.
     not check if a package extension indeed loads all of its trigger
     dependencies using `import` or `using`. 
 
-!!! warning "Known bug"
-
-    Currently, `Aqua.test_stale_deps` does not detect stale
-    dependencies when they are stdlib.  This is considered a bug and
-    may be fixed in the future.  Such a release is considered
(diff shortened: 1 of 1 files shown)


lgoettgens:
After creating a MWE and testing with different julia versions, `Serialization` gets reported as a stale dep in 1.11.0-alpha1 and nightly, but not in 1.10.2.
My guess is that this is due to the removal from the sysimg (https://github.com/JuliaLang/julia/pull/51399). Other stdlibs that are still in the sysimg (e.g. LinearAlgebra) still won't get detected.

Thus instead of removing the warning, I would change the "stdlib" to "package in the sysimage" or something.

LilithHafner:
I suspected that too, and even tested it to confirm. I must have made a mistake. I'll update the PR

lgoettgens (approved):
Thanks!

PR #313, opened 2024-11-24
feat: add test to detect public names without a docstring

### LINKED ISSUE #90: Test if all exported names has a docstring
we have some facilities testing this here: https://github.com/QuEraComputing/Bloqade.jl/pull/470/files#diff-8b6a3d1fcb6507d7155106fb9cbecf22d4b3b1377f4128af77b2112c2277e0a7R4

might be a good test for everyone in general? what do people think?

gdalle:
Fixes #90

Source:
- [x] Add a function `test_undocumented_names` which checks whether public names are documented or not (only on Julia 1.11 and later). It does so by verifying that `Docs.undocumented_names(module)` is either empty or contains only the module name itself.

Tests:
- [x] Add a kwarg `test_undocumented_names` to `test_all`
- [x] Add tests with one fully documented module and one partially documented module.

Docs:
- [x] Add docs page about `test_undocumented_names`

Chores:
- [ ] Bump version to v0.9.0 (BREAKING CHANGE)
- [x] Update CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Version [v0.9.0] - unreleased
+## Version [v0.8.12] - unreleased
 
 ### Changed
 
-- Add `test_undocumented_names` to verify that all public symbols have docstrings (not including the module itself). ([#313])
+- Add `test_undocumented_names` to verify that all public symbols have docstrings (not including the module itself). This test is not enabled by default in `test_all`. ([#313])
 
 ## Version [v0.8.11] - 2025-02-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Add `test_undocumented_names` to verify that all public symbols have docstrings (including the module itself). ([#313])
+- Add `test_undocumented_names` to verify that all public symbols have docstrings (not including the module itself). ([#313])
+
+## Version [v0.8.11] - 2025-02-06
+
+### Changed
+
+- Avoid deprecation warnings concerning `Base.isbindingresolved` with julia nightly. ([#322])
+
+## Version [v0.8.10] - 2025-01-26
+
+### Changed
+
+- No longer call `@testset` for testsets that are skipped. ([#319])
 
 
 ## Version [v0.8.9] - 2024-10-15
@@ -249,6 +261,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v0.8.7]: https://github.com/JuliaTesting/Aqua.jl/releases/tag/v0.8.7
 [v0.8.8]: https://github.com/JuliaTesting/Aqua.jl/releases/tag/v0.8.8
 [v0.8.9]: https://github.com/JuliaTesting/Aqua.jl/releases/tag/v0.8.9
+[v0.8.10]: https://github.com/JuliaTesting/Aqua.jl/releases/tag/v0.8.10
+[v0.8.11]: https://github.com/JuliaTesting/Aqua.jl/releases/tag/v0.8.11
 [#93]: https://github.com/JuliaTesting/Aqua.jl/issues/93
 [#103]: https://github.com/JuliaTesting/Aqua.jl/issues/103
 [#113]: https://github.com/JuliaTesting/Aqua.jl/issues/113
@@ -298,3 +312,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#288]: https://github.com/JuliaTesting/Aqua.jl/issues/288
 [#297]: https://github.com/JuliaTesting/Aqua.jl/issues/297
 [#309]: https://github.com/JuliaTesting/Aqua.jl/issues/309
+[#319]: https://github.com/JuliaTesting/Aqua.jl/issues/319
+[#322]: https://github.com/JuliaTesting/Aqua.jl/issues/322
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Version [v0.8.10] - unreleased
+
+### Changed
+
+- Add `test_undocumented_names` to verify that all public symbols have docstrings (including the module itself). ([#313])
+
 
 ## Version [v0.8.9] - 2024-10-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Version [v0.8.10] - unreleased
+## Version [v0.9.0] - unreleased
 
 ### Changed
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Aqua"
 uuid = "4c88cf16-eb10-579e-8560-4a9242c79595"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
-version = "0.9.0"
+version = "0.8.11"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Aqua"
 uuid = "4c88cf16-eb10-579e-8560-4a9242c79595"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
-version = "0.8.10"
+version = "0.9.0"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Aqua"
 uuid = "4c88cf16-eb10-579e-8560-4a9242c79595"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
-version = "0.8.9"
+version = "0.8.10"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -11,10 +11,6 @@ Aqua.jl provides functions to run a few automatable checks for Julia packages:
 * There are no "obvious" type piracies.
 * The package does not create any persistent Tasks that might block precompilation of dependencies.
 
-```@docs
-Aqua
-```
-
 ## Quick usage
 
 Call `Aqua.test_all(YourPackage)` from the REPL, e.g.,
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -11,6 +11,10 @@ Aqua.jl provides functions to run a few automatable checks for Julia packages:
 * There are no "obvious" type piracies.
 * The package does not create any persistent Tasks that might block precompilation of dependencies.
 
+```@docs
+Aqua
+```
+
 ## Quick usage
 
 Call `Aqua.test_all(YourPackage)` from the REPL, e.g.,
--- a/src/Aqua.jl
+++ b/src/Aqua.jl
@@ -68,44 +68,44 @@ function test_all(
     persistent_tasks = true,
     undocumented_names = true,
 )
-    @testset "Method ambiguity" begin
-        if ambiguities !== false
+    if ambiguities !== false
+        @testset "Method ambiguity" begin
             test_ambiguities([testtarget]; askwargs(ambiguities)...)
         end
     end
-    @testset "Unbound type parameters" begin
-        if unbound_args !== false
+    if unbound_args !== false
+        @testset "Unbound type parameters" begin
             test_unbound_args(testtarget; askwargs(unbound_args)...)
         end
     end
-    @testset "Undefined exports" begin
-        if undefined_exports !== false
+    if undefined_exports !== false
+        @testset "Undefined exports" begin
             test_undefined_exports(testtarget; askwargs(undefined_exports)...)
         end
     end
-    @testset "Compare Project.toml and test/Project.toml" begin
-        if project_extras !== false
+    if project_extras !== false
+        @testset "Compare Project.toml and test/Project.toml" begin
             isempty(askwargs(project_extras)) || error("Keyword arguments not supported")
             test_project_extras(testtarget)
         end
     end
-    @testset "Stale dependencies" begin
-        if stale_deps !== false
+    if stale_deps !== false
+        @testset "Stale dependencies" begin
             test_stale_deps(testtarget; askwargs(stale_deps)...)
         end
     end
-    @testset "Compat bounds" begin
-        if deps_compat !== false
+    if deps_compat !== false
+        @testset "Compat bounds" begin
             test_deps_compat(testtarget; askwargs(deps_compat)...)
         end
     end
-    @testset "Piracy" begin
-        if piracies !== false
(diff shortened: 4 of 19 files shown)


src/undocumented_names.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    test_undocumented_names(module::Module)
+
+Test that all public names in `module` have a docstring (including the module itself).
```
ericphanson, line 4: Imo the module itself should be exempt bc it has an auto-generated docstring (the readme) which is often more useful than whatever short thing folks will type in to pass the aqua check

src/Aqua.jl
```diff
@@ -1,6 +1,13 @@
+"""
+    Aqua
```
ericphanson, line 2: This docstring seems less useful than the readme to me

docs/src/index.md
```diff
@@ -11,6 +11,10 @@ Aqua.jl provides functions to run a few automatable checks for Julia packages:
 * There are no "obvious" type piracies.
 * The package does not create any persistent Tasks that might block precompilation of dependencies.
 
+```@docs
+Aqua
```
KeithWM, line 15: As already pointed out, this is indeed a bit of a useless docstring to add. If typically packages do not have a docstring, which I think they indeed do not, it will hamper adoption of this feature if we impose it. Worse still, it might hurt Aqua as a whole if this test becomes a default or limit the chances of this test becoming a default.

src/undocumented_names.jl
gdalle: Sounds good, I'll make the change.

Out of curiosity, how does the README go into an auto-generated docstring here? I know DocStringExtensions.jl can do that but is there another mechanism I'm not aware of?

src/Aqua.jl
gdalle: I'll remove it and adapt the test

src/Aqua.jl
```diff
@@ -105,6 +116,13 @@ function test_all(
             test_persistent_tasks(testtarget; askwargs(persistent_tasks)...)
         end
     end
+    @testset "Undocumented names" begin
+        if undocumented_names !== false
```
KeithWM, line 113: It looks to me that this Codecov warning is not accurate, right?

src/undocumented_names.jl
```diff
@@ -0,0 +1,26 @@
...
+
+!!! warning
+    For Julia versions before 1.11, this does not test anything.
+"""
```
KeithWM, line 8: I would prefer seeing this warning in code with an `@warn` rather than a comment.

test/test_undocumented_names.jl
```diff
@@ -0,0 +1,29 @@
+module TestUndocumentedNames
+
+include("preamble.jl")
+
+import PkgWithUndocumentedNames
+import PkgWithoutUndocumentedNames
+
+# Pass
+results = @testtestset begin
+    Aqua.test_undocumented_names(PkgWithoutUndocumentedNames)
```
KeithWM, line 10: I would expect an `@testset` `begin`-`end` block to be indented, or am I mistaken?

src/Aqua.jl
```diff
@@ -63,6 +73,7 @@ function test_all(
     deps_compat = true,
     piracies = true,
     persistent_tasks = true,
+    undocumented_names = true,
```
KeithWM, line 69: Setting this to true implies that dependents of this package might now fail their tests, right? In that case, shouldn't this be a breaking change, i.e. up to `v0.9.1` instead of `v0.8.10`?

KeithWM (commented):
Just some comments. I am not an experienced enough Julia developer to be entirely confident that everything I say is correct, but I hope you value the suggestions/questions.

docs/src/index.md
gdalle: Yes, I'll remove that requirement.

src/Aqua.jl
gdalle: indeed, Codecov misses the lines that start with `@testset`

src/undocumented_names.jl
gdalle: That means adding a warning for every test below Julia 1.11, not sure people will like that. The fact that it does nothing on 1.10 can already be seen from the fact that the relevant `@testset` is empty

test/test_undocumented_names.jl
gdalle: There doesn't have to be one, the loop in `runtests.jl` already wraps each file into a `@testset`

src/Aqua.jl
gdalle: I don't know what Aqua's versioning policy is when adding a new test. I'll defer to the maintainers

src/undocumented_names.jl
ericphanson: Ian added it to 1.7: https://github.com/JuliaLang/julia/pull/39093

gdalle: nice!

ericphanson (commented):
the feature and implementation look good to me, but I think it is breaking to add a new default-on check (both on the principle that it would be very annoying to start failing CI "out of nowhere" and precedent, e.g. https://github.com/JuliaTesting/Aqua.jl/pull/174).

I think this could start off-by-default or just be v0.9.0.

src/undocumented_names.jl
```diff
@@ -0,0 +1,26 @@
...
+    test_undocumented_names(module::Module)
+
+Test that all public names in `module` have a docstring (not including the module itself).
+
+!!! warning
+    For Julia versions before 1.11, this does not test anything.
+"""
+function test_undocumented_names(m::Module)
+    @static if VERSION >= v"1.11"
+        undocumented_names = filter(n -> n != nameof(m), Docs.undocumented_names(m))
```
stevengj, line 11:
```suggestion
        undocumented_names = filter(n -> n != nameof(m), Docs.undocumented_names(m))
```
Why the filter?   It's good practice for a module to have a docstring.

ericphanson: context: https://github.com/JuliaTesting/Aqua.jl/pull/313#discussion_r1855426913

gdalle:
> I think this could start off-by-default or just be v0.9.0.

Either is fine by me. I guess it depends whether there are other breaking changes which might benefit from tagging v0.9.0.

ericphanson:
Personally I think doing a breaking release whenever there's a new check is alright, and just have all checks opt-out by default. I'm not an Aqua maintainer, but I've used aqua in a lot of repos with varying levels of activity, and I haven't found breaking releases of Aqua particularly annoying. No other packages (should) depend on Aqua so being on an old release doesn't hold back other packages, and it's fine to just stay on an older release until the next maintenance pass. I think mixing opt-in and opt-out is a bit more confusing and requires more attention for users.

gdalle:
Tentatively bumping the version to v0.9.0, we'll see what the maintainers thin

src/undocumented_names.jl
KeithWM: I realize now that this implementation relies heavily on the "new" `public` keyword, which was not actually in the original issue (as it predated the keyword). Given that a main reason for introducing the `public` keyword was that previously something being documented was used to show it was part of the API, **I think the proposed implementation is a good idea**. But I'm not quite sure how to deal with the expectations towards <1.11 users. It _could_ help to change the name of the check to something like `test_public_documentation`, `test_documented_api` or something else that include the word "public" without being as verbose as `test_undocumented_public_names`.

KeithWM:
> Tentatively bumping the version to v0.9.0, we'll see what the maintainers think

I think the whole point of semver falls to bits if packages start cutting corners on what is and what is not breaking. I'm not sure what I usually see happening in practice, what kind of compat (caret, fixed, free) package developers have with their Aqua (test) dependency. A suggestion might be that Aqua warns users of newer versions, without actually breaking anything. Or that the documentation recommends dependents do not use a compat for Aqua, to avoid Aqua-clearance giving a false sense of security that all is well.

KeithWM (approved):
I'm happy with this, but I wouldn't take my approval as a sufficient proof that the PR is good.

src/undocumented_names.jl
gdalle: Given that `Docs.undocumented_names` is not defined on 1.10, I'm not sure we can do anything meaningful before 1.11 (unless that function is added to Compat.jl)

j-fu:
LGTM

gdalle:
@lgoettgens would you mind taking a look? I got several reviews but none of them from maintainers I think?

gdalle:
gentle bump on this one, perhaps @fingolfin would be open to reviewing?

ufechner7:
Any update?

gdalle:
@lgoettgens from my end this is good to go

fingolfin:
Sorry for the long time.

Needs a rebase in the meantime :-/ but hopefully an easy one.

src/undocumented_names.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    test_undocumented_names(module::Module)
+
+Test that all public names in `module` have a docstring (not including the module itself).
+
+!!! warning
+    For Julia versions before 1.11, this does not test anything.
```
fingolfin, line 7: Maybe a bit clearer:
```suggestion
    When running this Aqua test in Julia versions before 1.11, it does nothing.
```
or even more explicit:
```suggestion
    When running this Aqua test in Julia versions before 1.11, it does nothing.
    Thus if you use continuous integration tests, make sure those are configured
    to use Julia >= 1.11 in order to benefit from this test.
```

src/undocumented_names.jl
fingolfin: Ideally that context would be added as a comment here, so that future readers of the code also have a chance to know it :-)

src/undocumented_names.jl
fingolfin: Hmm, let me think loudly about this a bit ....

If we are concerned about people accidentally not running this test because they have their CI configured to run Aqua only with, say, Julia 1.10 (as LTS release), then it would indeed make sense to be more aggressive about flagging this. But I am not sure a `@warn` is that much more likely to be seen, esp. in a CI runner. If we want to ensure that, we'd ought to raise an error that suggests two options: either run Aqua in Julia >= 1.11, or else (for those of us who run tests in multiple Julia versions, like e.g. most projects I am involved in do) use a workaround, like 
> ... pass the option `undocumented_names = VERSION >= v"1.11"` to `Aqua.test_all`.

Alas, we have not done that for other tests. In the end, I don't think this inconvenience (be it a warning or an error) for most people is worth it or necessary: the problem will solve itself over time, either by projects eventually switching to newer Julia versions, at which point they'll get the test; or by projects being ended and thus never getting this, but also never needing it.


That leaves the people who desperately try to get `undocumented_names = true` to do something but somehow missed that their Julia is to old. For them a `@warn` might be helpful.

To restrict it to those folks, we could also change the default for `undocumented_names` to something like this: `undocumented_names = VERSION >= v"1.11"`, and then change `test_undocumented_names` to raise an error (or at least a warning) if it is called in Julia < 1.11.

This way, anyone using `test_all` with default arguments gets the test in >= 1.11 and otherwise not, without a warning; but someone who desperately tries to enable the test by passing `test_undocumented_names = true` will get a helpful message as to why the test is not turning on despite their efforts.



Anyway: I am also perfectly fine with the code as it is.

fingolfin (changes requested):
Thank you very much for this PR, and so sorry for not tending to it for so long.

Needs a rebase, and a few things should be clarified in the documentation. Otherwise it seems good.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Version [v0.9.0] - unreleased
+
+### Changed
+
+- Add `test_undocumented_names` to verify that all public symbols have docstrings (including the module itself). ([#313])
```
lgoettgens, line 12:
```suggestion
- Add `test_undocumented_names` to verify that all public symbols have docstrings (not including the module itself). ([#313])
```
I think this was forgotten in 4df5e8b62b561553784547abae2d78f9c24dc4ef

src/undocumented_names.jl
gdalle: I think we can merge this as it is, and then possibly add a warning on 1.10 in a later release.

fingolfin (approved):
Looks good to me

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Version [v0.9.0] - unreleased
+
+### Changed
+
+- Add `test_undocumented_names` to verify that all public symbols have docstrings (not including the module itself). ([#313])
```
lgoettgens, line 12:
```suggestion
## Version [v0.8.12] - unreleased

### Changed

- Add `test_undocumented_names` to verify that all public symbols have docstrings (not including the module itself). This test is not enabled by default in `test_all`. ([#313])
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Aqua"
 uuid = "4c88cf16-eb10-579e-8560-4a9242c79595"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
-version = "0.8.11"
+version = "0.9.0"
```
lgoettgens, line 4:
```suggestion
version = "0.8.11"
```

src/Aqua.jl
```diff
@@ -52,6 +54,7 @@ passed to `\$x` to specify the keyword arguments for `test_\$x`.
 - `deps_compat = true`
 - `piracies = true`
 - `persistent_tasks = true`
+- `undocumented_names = true`
```
lgoettgens, line 57:
```suggestion
- `undocumented_names = false`
```

src/Aqua.jl
```diff
@@ -63,6 +66,7 @@ function test_all(
     deps_compat = true,
     piracies = true,
     persistent_tasks = true,
+    undocumented_names = true,
```
lgoettgens, line 69:
```suggestion
    undocumented_names = false,
```

lgoettgens (approved):
I would like to release this as a patch release with this test disabled by default to not disrupt the whole ecosystem, as there are many packages that don't declare a compat bound on Aqua. Thus, I'll apply the few suggestions here and will go ahead with the release later today

src/undocumented_names.jl
```diff
@@ -0,0 +1,33 @@
...
+
+!!! warning
+    When running this Aqua test in Julia versions before 1.11, it does nothing.
+    Thus if you use continuous integration tests, make sure those are configured
+    to use Julia >= 1.11 in order to benefit from this test.
+"""
+function test_undocumented_names(m::Module)
+    @static if VERSION >= v"1.11"
+        # exclude the module name itself because it has the README as auto-generated docstring (https://github.com/JuliaLang/julia/pull/39093)
+        undocumented_names = filter(n -> n != nameof(m), Docs.undocumented_names(m))
```
lgoettgens, line 18: When having a final look at this, I stumbled over something:

IMO, this should iterate over submodules as well, but only over submodules that are part of the same package. There should be some helper functions to do that already somewhere in Aqua (e.g. https://github.com/JuliaTesting/Aqua.jl/blob/b7d973fb12e9d324441fc54da7455698acffe83b/src/exports.jl#L26-L30)

lgoettgens (changes requested):

gdalle:
> I would like to release this as a patch release with this test disabled by default to not disrupt the whole ecosystem, as there are many packages that don't declare a compat bound on Aqua.

What is the difference? One day there will be a breaking release switching this on, and people without a compat bound on Aqua will get surprised in exactly the same way.

Maybe we could add a warning telling people to switch to `true` for this setting, that way at least they catch it if they look at the test output? Otherwise I'm not sure anyone is actively monitoring release notes for Aqua's patch releases.

src/undocumented_names.jl
```diff
@@ -0,0 +1,33 @@
...
+!!! tip
+    On all Julia versions, public names include the exported names.
+    On Julia versions >= 1.11, public names also include the names annotated with the `public` keyword.
+
+!!! warning
+    When running this Aqua test in Julia versions before 1.11, it does nothing.
+    Thus if you use continuous integration tests, make sure those are configured
+    to use Julia >= 1.11 in order to benefit from this test.
+"""
+function test_undocumented_names(m::Module)
```
lgoettgens, line 15: This should also accept a `broken` kwarg, with the same semantics as for e.g. `test_undefined_exports`

gdalle: done in the latest commit

src/undocumented_names.jl
gdalle: done in the latest commit

lgoettgens (approved):

PR #316, opened 2025-01-08
test_unbound_args: add a functionality to ignore specific methods

arnaud-ma:
With this PR, it is now possible to ignore some methods, given their signature. For example, taking the false positive in #86: 
```julia
module abcd

f(::NTuple{N, T}) where {N,T} = (N,T)
f(::Tuple{}) = (0,Any)

end # module
```

It is possible to disable the false positive with:
```julia
Aqua.test_unbound_args(abcd, ignore = [(abcd.f, NTuple)])
```

This specific example has been added in the unit tests.

~To improve flexibility, the possibility to call directly `Aqua.test_unbound_args(unbounds_args)` where `unbounds_args` is a collection of methods (probably got with `Test.detect_unbound_args`) is also added.~

### DIFF
--- a/src/unbound_args.jl
+++ b/src/unbound_args.jl
@@ -11,19 +11,19 @@ passed directly to the function instead of the module.
 # Keyword Arguments
 - `broken::Bool = false`: If true, it uses `@test_broken` instead of
   `@test` and shortens the error message.
-- `ignore::Vector{Tuple{Function, DataType...}}`: A list of functions and their
-  signatures to ignore. The signatures are given as tuples, where the
-  first element is the function and the rest are the types of the
-  arguments. For example, to ignore `foo(x::Int, y::Float64)`, pass
-  `(foo, Int, Float64)`.
+- `exclude::AbstractVector{Tuple{Function, DataType...}} = []`: A list of
+  functions and their signatures to exclude. The signatures are given as
+  tuples, where the first element is the function and the rest are the types of
+  the arguments. For example, to ignore `foo(x::Int, y::Float64)`,
+  pass `(foo, Int, Float64)`.
 """
-function test_unbound_args(m::Module; broken::Bool = false, ignore = ())
+function test_unbound_args(m::Module; broken::Bool = false, exclude = [])
     unbounds = detect_unbound_args_recursively(m)
-    for i in ignore
+    for i in exclude
         # i[2:end] is empty if length(i) == 1
-        ignore_signature = Tuple{typeof(i[1]),i[2:end]...}
+        exclude_signature = Tuple{typeof(i[1]),i[2:end]...}
         filter!(unbounds) do method
-            method.sig != ignore_signature
+            method.sig != exclude_signature
         end
     end
     test_unbound_args(unbounds; broken = broken)
--- a/src/unbound_args.jl
+++ b/src/unbound_args.jl
@@ -20,8 +20,11 @@ passed directly to the function instead of the module.
 function test_unbound_args(m::Module; broken::Bool = false, exclude = [])
     unbounds = detect_unbound_args_recursively(m)
     for i in exclude
-        # i[2:end] is empty if length(i) == 1
-        exclude_signature = Tuple{typeof(i[1]),i[2:end]...}
+        callable, args = i[1], i[2:end] # i[2:end] is empty if length(i) == 1
+
+        # the type of the function is the function itself if it is a callable object
+        callable_t = callable isa Function ? typeof(callable) : callable
+        exclude_signature = Tuple{callable_t, args...}
         filter!(unbounds) do method
             method.sig != exclude_signature
         end
--- a/src/unbound_args.jl
+++ b/src/unbound_args.jl
@@ -1,17 +1,35 @@
 """
     test_unbound_args(module::Module)
+    test_unbound_args(unbounds)
 
 Test that all methods in `module` and its submodules do not have
 unbound type parameters. An unbound type parameter is a type parameter
 with a `where`, that does not occur in the signature of some dispatch
-of the method.
+of the method. If unbounds methods are already known, they can be
+passed directly to the function instead of the module.
 
 # Keyword Arguments
 - `broken::Bool = false`: If true, it uses `@test_broken` instead of
   `@test` and shortens the error message.
+- `ignore::Vector{Tuple{Function, DataType...}}`: A list of functions and their
+  signatures to ignore. The signatures are given as tuples, where the
+  first element is the function and the rest are the types of the
+  arguments. For example, to ignore `foo(x::Int, y::Float64)`, pass
+  `(foo, Int, Float64)`.
 """
-function test_unbound_args(m::Module; broken::Bool = false)
+function test_unbound_args(m::Module; broken::Bool = false, ignore = ())
     unbounds = detect_unbound_args_recursively(m)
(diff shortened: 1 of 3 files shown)


cgarling:
The ability to ignore specific methods when testing for unbound arguments would be helpful.

src/unbound_args.jl
```diff
@@ -1,17 +1,35 @@
 """
     test_unbound_args(module::Module)
+    test_unbound_args(unbounds)
 
 Test that all methods in `module` and its submodules do not have
 unbound type parameters. An unbound type parameter is a type parameter
 with a `where`, that does not occur in the signature of some dispatch
-of the method.
+of the method. If unbounds methods are already known, they can be
+passed directly to the function instead of the module.
```
lgoettgens, line 9: I don't quite understand what this second dispatch should be used for? It is basically just a `@test isempty` with some extra printing. 

Furthermore, the dispatch does not honour the same kwargs, which would need to be pointed out in the docstring. But I am more in favor of removing this dispatch altogether.

src/unbound_args.jl
```diff
@@ -1,17 +1,35 @@
...
+  first element is the function and the rest are the types of the
+  arguments. For example, to ignore `foo(x::Int, y::Float64)`, pass
+  `(foo, Int, Float64)`.
 """
-function test_unbound_args(m::Module; broken::Bool = false)
+function test_unbound_args(m::Module; broken::Bool = false, ignore = ())
     unbounds = detect_unbound_args_recursively(m)
+    for i in ignore
+        # i[2:end] is empty if length(i) == 1
+        ignore_signature = Tuple{typeof(i[1]),i[2:end]...}
```
lgoettgens, line 24: This approach unfortunately does not work for callable objects.

Please make sure that this also works for cases like
```julia
struct S{U}
  s::U
end

(::S{U})(::NTuple{N,T}) where {N,T,U} = (N,T,U)
(::S{U})(::Tuple{}) where {U} = (0,Any,U)
``` 
and include such an example in the tests

src/unbound_args.jl
```diff
@@ -1,17 +1,35 @@
...
 unbound type parameters. An unbound type parameter is a type parameter
 with a `where`, that does not occur in the signature of some dispatch
-of the method.
+of the method. If unbounds methods are already known, they can be
+passed directly to the function instead of the module.
 
 # Keyword Arguments
 - `broken::Bool = false`: If true, it uses `@test_broken` instead of
   `@test` and shortens the error message.
+- `ignore::Vector{Tuple{Function, DataType...}}`: A list of functions and their
```
lgoettgens, line 14: For the ambiguity test, the analogous kwarg there is named `exclude`. I would prefer for these to be more or less aligned.

lgoettgens (commented):
Thanks for working on this. Adding exceptions is indeed something that each of the test functions should have. Some detailed comments below

src/unbound_args.jl
arnaud-ma: This was to give more flexibility. The second dispatch was exactly what test_unbound_args did before, except that `unbounds` is a parameter instead of being hardcoded with the value of `Test.detect_unbound_args(module)`. But it's true that now that there is `exclude`, the uses are very limited and probably dangerous. It's probably safer to hardcode `unbounds`.

arnaud-ma: This was to give more flexibility. The second dispatch was exactly what test_unbound_args did before, except that `unbounds` is a parameter instead of being hardcoded with the value of `Test.detect_unbound_args(module)`. But it's true that now that there is `exclude`, the uses are very limited and probably dangerous. It's probably safer to hardcode `unbounds`.

arnaud-ma:
All done!

src/unbound_args.jl
```diff
@@ -9,9 +9,23 @@ of the method.
...
+  pass `(foo, Int, Float64)`.
 """
-function test_unbound_args(m::Module; broken::Bool = false)
+function test_unbound_args(m::Module; broken::Bool = false, exclude = [])
     unbounds = detect_unbound_args_recursively(m)
+    for i in exclude
+        callable, args = i[1], i[2:end] # i[2:end] is empty if length(i) == 1
+
+        # the type of the function is the function itself if it is a callable object
+        callable_t = callable isa Function ? typeof(callable) : callable
```
lgoettgens, line 24: Hmm, I thinks this still isn't quite what we want here.
For custom structs, say `struct S end`, there are both function signatures containing `S` and `Type{S}` (which isn't even quite `typeof(S)`), since the former is for callable objects and the latter for constructors.

Furthermore, there are subtypes of `Function`, that are not a `singleton type of function xxx`, e.g. `Base.ComposedFunction`.

Maybe you have an idea that makes all of these cases work. If not, I think we just need the user to specify the exact signature, i.e. `foo(x::Int, y::Float64)` as `(typeof(foo), Int, Float64)`.

arnaud-ma: Ah, I understand now. Thank you very much for your patience. After some research, I can't find any way to be 100% sure that we are dealing with a "simple" method.

But specifying the exact signature requires some knowledge from the user. To help the user, I think adding what method.sig returns in the error message can be a good idea. For example from this:
```diff
Unbound type parameters detected:
[1] f(x) where T @ Main.Foo ~/projets/forks/Aqua/Aqua.jl/a.jl:15
[2] (::Main.Foo.S{U})(::NTuple{N, T}) where {N, T, U} @ Main.Foo ~/projets/forks/Aqua/Aqua.jl/a.jl:8
[3] Main.Foo.S(::NTuple{N, T}) where {N, T} @ Main.Foo ~/projets/forks/Aqua/Aqua.jl/a.jl:11
```

to this:
```
Unbound type parameters detected:

[1] f(x) where T @ Main.Foo ~/projets/forks/Aqua/Aqua.jl/a.jl:15
    signature: Tuple{typeof(Main.Foo.f), Any} where T

[2] (::Main.Foo.S{U})(::NTuple{N, T}) where {N, T, U} @ Main.Foo ~/projets/forks/Aqua/Aqua.jl/a.jl:8
    signature: Tuple{Main.Foo.S{U}, NTuple{N, T}} where {N, T, U}

[3] Main.Foo.S(::NTuple{N, T}) where {N, T} @ Main.Foo ~/projets/forks/Aqua/Aqua.jl/a.jl:11
    signature: Tuple{Type{Main.Foo.S}, NTuple{N, T}} where {N, T}
```

What do you think about that?

lgoettgens: I am not a fan of requiring 3x of vertical space. Instead, I think we could make a function public that returns the list of methods, so that a user can query this for the signature themselves. But I would put that to a follow-up PR.
For this PR, it would be great if you could do the small adaptions needed for this, and adapt the docstring as well.

jlapeyre:
If you are willing to rewrite your code, this may be a better workaround:
https://discourse.julialang.org/t/is-this-test-detect-unbound-args-result-valid-or-a-bug/96987


The Aqua documentation on `test_unbound_args` might also mention this. The fix recommended in the docs is not applicable in most situations.

```julia
julia> f(::Tuple{T, Vararg{T, NmOne}}) where {T, NmOne} = (NmOne + 1 ,T);

julia> f((1,2,3))
(3, Int64)

julia> import Test;

julia> Test.detect_unbound_args(Main)
0-element Vector{Method}
```

PR #319, opened 2025-01-23
Don't run the `@testset` if there are no tests

kuszmaul:
Some testing tools complain if a `@testset` is empty.  So don't run empty test sets.

### DIFF
--- a/src/Aqua.jl
+++ b/src/Aqua.jl
@@ -91,7 +91,7 @@ function test_all(
         end
     end
     if deps_compat !== false
-    @testset "Compat bounds" begin
+        @testset "Compat bounds" begin
             test_deps_compat(testtarget; askwargs(deps_compat)...)
         end
     end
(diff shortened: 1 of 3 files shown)


src/Aqua.jl
```diff
@@ -64,44 +64,44 @@ function test_all(
...
     end
-    @testset "Stale dependencies" begin
-        if stale_deps !== false
+    if stale_deps !== false
+        @testset "Stale dependencies" begin
             test_stale_deps(testtarget; askwargs(stale_deps)...)
         end
     end
+    if deps_compat !== false
     @testset "Compat bounds" begin
```
nickrobinson251, line 94: missing indent here?

nickrobinson251 (approved):
LGTM

But ideally a maintainer of this package should review / merge

src/Aqua.jl
kuszmaul: fixed

lgoettgens (approved):
It indeed made little sense to print the empty testsets for testsets that were intentionally skipped. Thanks

PR #327, opened 2025-04-28
use more precise structural inspection of types for ambiguity exclusions

vtjnash:
Should be slightly more robust, since subtyping already implements these queries with much more finesse than is going to be true of a re-implementation of the query here.

### DIFF
--- a/test/test_ambiguities.jl
+++ b/test/test_ambiguities.jl
@@ -49,7 +49,7 @@ include("preamble.jl")
     check_testcase([PkgWithAmbiguities.ConcreteType], total - num_ambs_ConcreteType)
 
     # exclude abstract supertype without callables and constructors
-    check_testcase([PkgWithAmbiguities.AbstractType], total - num_ambs_SingletonType - num_ambs_ConcreteType)
+    check_testcase([PkgWithAmbiguities.AbstractType], total)
 
     # for ambiguities between abstract and concrete type callables, only one needs to be excluded
     check_testcase(
--- a/src/ambiguities.jl
+++ b/src/ambiguities.jl
@@ -29,30 +29,31 @@ test_ambiguities(packages; kwargs...) = _test_ambiguities(aspkgids(packages); kw
 
 const ExcludeSpec = Pair{Base.PkgId,String}
 
-strnameof(x) = string(x)
-strnameof(x::Type) = string(nameof(x))
-
-rootmodule(x) = rootmodule(parentmodule(x))
+rootmodule(x::Type) = rootmodule(parentmodule(x))
 rootmodule(m::Module) = Base.require(PkgId(m))  # this handles Base/Core well
 
-normalize_exclude(x::Union{Type,Function}) =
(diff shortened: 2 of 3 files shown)


vtjnash:
Do you still care about v1.2- here? Apparently it needs some fix to avoid a julia bug, but I don't know if that is actually helpful to figure out?

vtjnash:
I can also fix the support for those versions, I just wanted to check first if this PR and that support was considered useful

test/test_ambiguities.jl
```diff
@@ -30,6 +30,11 @@ include("preamble.jl")
         if broken
             @test_broken num_ambiguities_ == num_ambiguities
         else
+            if num_ambiguities_ != num_ambiguities
+                @show exclude
+                println(strout)
+                println(strerr)
+            end
```
fingolfin, line 37: How is this related to the rest of the changes in here?

fingolfin (approved):
Sounds like a good idea, thank you!

test/test_ambiguities.jl
vtjnash: Just debugging for the relevant tests if they change reporting

test/test_ambiguities.jl
```diff
@@ -44,7 +49,7 @@ include("preamble.jl")
     check_testcase([PkgWithAmbiguities.ConcreteType], total - num_ambs_ConcreteType)
 
     # exclude abstract supertype without callables and constructors
-    check_testcase([PkgWithAmbiguities.AbstractType], total)
+    check_testcase([PkgWithAmbiguities.AbstractType], total - num_ambs_SingletonType - num_ambs_ConcreteType)
```
lgoettgens, line 52: This seems like an unwanted behavioral change.
If I exclude `AbstractType`, according to the docstring, this excludes ambiguities of the constructor `AbstractType` and the functor `(::AbstractType)`.
However, after this change, it also ignores ambiguities of the constructor `SingletonType` (which IMO is a completely different function) and the functor `(::SingletonType)` (which shares the methodtable with `(::AbstractType)` for each `SingletonType` object created).
The former change seems clearly unwanted to me, for the latter one I am not sure what to do best here. If possible, I think I would like to keep the current behavior.

Note, however, that with the current behavior, excluding `ConcreteParameterizedType` also excludes all constructors `ConcreteParameterizedType{T}` for any `T`.

Since we already have a breaking change in master, it would be fine to slightly change the semantics of exclusion here, but that would need some good explanation of the new semantics in the docstring of `test_ambiguities` and in the changelog.

lgoettgens (changes requested):

test/test_ambiguities.jl
vtjnash: Okay, I've switched it back to doing only structural inspection instead of the subtyping queries and just simplified them to stay in the type domain where the comparisons can be made more reliable. It is less powerful (since it cannot realize that `SingletonType` is an implementation of `AbstractType`) but that keeps the existing comparison behavior.

PR #332, opened 2025-05-08
Don't create testset if no test to run

nickrobinson251:
same as https://github.com/JuliaTesting/Aqua.jl/pull/319 but for the newly added test from https://github.com/JuliaTesting/Aqua.jl/pull/313


lgoettgens (approved):
Thanks! I must have missed that in the review

PR #334, opened 2025-05-27
update piracies code to support v1.12 changes

vtjnash:
Reflects current status of JuliaLang/julia#58131 PR to remove unnecessary work that is no longer viable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Adapt to internal method table changes in Julia 1.12 and later. ([#334])
 - The minimum supported julia version is increased to 1.6. ([#328])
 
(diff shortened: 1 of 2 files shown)


lgoettgens:
Is https://github.com/JuliaLang/julia/pull/58131 still going into 1.12? I am asking this surprised as it is already quite late in the release cycle and that PR doesn't have a backport label.

I'll give this a merge once there is a prerelease/nightly build that verifies that in indeed works in action.

But for right now, all CI failed due to something with `Base.visit`.

lgoettgens (approved):
LGTM, Thanks! I'll re-run CI and merge this once there is a nightly version containing https://github.com/JuliaLang/julia/pull/58131 (in the next two days or so). That way, this patch should already be available when 1.12.0-beta3 gets released

lgoettgens:
The current CI run uses `Julia Version 1.13.0-DEV.647` which already includes https://github.com/JuliaLang/julia/pull/58131.

Issue #310, opened 2024-10-15
1.0 Roadmap


lgoettgens:
This issue contains a list of things that have to be done/changed in Aqua before I would call out a 1.0 release.
@ everyone feel free to comment your opinions to this below.

- [ ] docs: versioning strategy (x-ref https://github.com/JuliaTesting/Aqua.jl/issues/229)
- [ ] reduce code duplication in `test_all`; instead have some registering mechanism
- [ ] split different tests into submodules (x-ref https://github.com/JuliaTesting/Aqua.jl/pull/154)
- [ ] docs: how to add a new test
- go through the already existing tests, clean them up, resolve major issues:
  - [ ] `test_ambiguities`: 
    - [x] https://github.com/JuliaTesting/Aqua.jl/issues/77
    - [ ] https://github.com/JuliaTesting/Aqua.jl/issues/308
    - [ ] https://github.com/JuliaTesting/Aqua.jl/issues/290
    - [ ] make it somehow work with package extensions
    - [ ] re-work interface (more ways to interact than just the one `test_*` function)
  - [ ] `test_unbound_args`:
    - [ ] make it somehow work with package extensions
    - [ ] re-work interface (more ways to interact than just the one `test_*` function)
  - [ ] `test_undefined_exports`:
    - [ ] re-work interface (more ways to interact than just the one `test_*` function)
  - [ ] `test_stale_deps`:
    - [ ] https://github.com/JuliaTesting/Aqua.jl/issues/264
    - [ ] re-work interface (more ways to interact than just the one `test_*` function)
  - [ ] `test_project_extras`:
    - [ ] https://github.com/JuliaTesting/Aqua.jl/issues/251
    - [ ] Decide if this test is still needed.
  - [ ] `test_deps_compat`:
    - [ ] https://github.com/JuliaTesting/Aqua.jl/issues/294
  - [ ] `test_piracies`:
    - [ ] re-work interface (more ways to interact than just the one `test_*` function)
    - [ ] make it somehow work with package extensions
  - [ ] `test_persistent_tasks`:
    - [ ] https://github.com/JuliaTesting/Aqua.jl/issues/292

New tests: Since they have not been mentioned in the above list, I would like to add some things about them. Until the process to 1.0 is finished, I won't put any effort myself into adding new tests. If someone opens a PR with a new test, this will probably need to wait for some time where it can be incorporated in a between-version of this process, or even until 1.0 is finished, and then added as described in the versioning strategy.

[...]

PR #350, opened 2025-08-21
remove unused dependency on Compat.jl

ranocha:
This dependency has already been removed in #221. It was added again in #224. However, it has not been removed when finally updating the required Julia version in #328.

This fixes an issue found after adding a Downgrade check to RootedTrees.jl in https://github.com/SciML/RootedTrees.jl/pull/192 and https://github.com/SciML/RootedTrees.jl/pull/194.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,12 +4,10 @@ authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
 version = "1.0.0-DEV"
 
 [deps]
-Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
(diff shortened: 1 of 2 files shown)


lgoettgens:
I don't understand how this should fix anything on any current release of Aqua. Iiuc, this should have only affected the master branch. For the 1.0 progress, see https://github.com/JuliaTesting/Aqua.jl/issues/310

ranocha:
The Downgrade CI jobs added to a few repositories fail because of this (see the links in the first post of this PR). Some dependencies seem to introduce old versions of Compat somewhere and we do not want to maintain different versions of Aqua.jl in the tests.

ranocha:
Well, it would be fine if we could use a new version of Aqua.jl that does not depend on Compat.jl anymore. That's what motivated me to create this PR.

PR #369, opened 2026-02-23
feat!: test imports and use of private APIs through ExplicitImports.jl

### LINKED ISSUE #268: Check unused explicit imports
One of my packages broke because of

```julia
using Base.Iterators: countfrom, repeat
```

As of Julia 1.9, `repeat` is no longer available that way, and I didn't notice that it was in Base all along.
While checking the other imported identifiers, I noticed some other candidates that didn't even need anymore.
It would be awesome if Aqua could detect those.

gdalle:
Fixes #268:

- [x] Add a new `Aqua.test_explicit_imports` which forwards everything to `ExplicitImports.test_explicit_imports` (see its [documentation](https://juliatesting.github.io/ExplicitImports.jl/stable/api/#Checks-to-use-in-testing))
- [x] Add two new dependencies to Aqua: ExplicitImports and TOML, clean up some other top-level imports
- [x] Bump Julia compat lower bound to 1.10, remove CI on versions older than LTS
- [x] Ignore enough private stuff for the new test to pass on Aqua itself
- [x] Add documentation that points towards the ExplicitImports docs

Main questions:

- Is it okay to add ExplicitImports as a hard dependency? It seems necessary if we want to make this test a default one, but it constraints Julia compat.
- Is it okay to just forward all kwargs to ExplicitImports? Or do we want to change the default values for some of them? Note that activating all of these tests will probably make most packages fail the Aqua checks (but that's arguably a good thing, and that's why we put it in the breaking v1 release).
- Is the smoke test on Aqua itself sufficient for this one, since it is properly tested in its native package, or do we want specific tests for other kwargs?

Ping @ericphanson

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,8 +32,6 @@ jobs:
           version: 1
       - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-docdeploy@v1
-        with:
-          install-package: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DOCUMENTER_KEY: ${{ secrets.SSH_KEY }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,6 +32,8 @@ jobs:
           version: 1
       - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-docdeploy@v1
+        with:
+          install-package: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DOCUMENTER_KEY: ${{ secrets.SSH_KEY }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest']
+        os: ["ubuntu-latest"]
         julia-version:
           - '1.12-nightly'
           - '1.11'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,21 +29,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest"]
+        os: ['ubuntu-latest']
         julia-version:
           - '1.12-nightly'
           - '1.11'
           - '1.10'
-          - '1.9'
-          - '1.8'
-          - '1.7'
-          - '1.6'
           - 'nightly'
         include:
           - os: windows-latest
             julia-version: '1'
-          - os: windows-latest
-            julia-version: '1.6'
           - os: windows-latest
             julia-version: '1.12-nightly'
           - os: windows-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The minimum supported julia version is increased to 1.6. ([#328])
+- Explicit imports are tested using ExplicitImports.jl ([#369])
 
 ## Version [v0.8.14] - 2025-08-04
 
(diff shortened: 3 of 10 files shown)


ericphanson:
My concern is not the breakage, it's that I think Aqua serves as a set of recommendations and I don't think it's the right recommendation to make in general:

- I don't think the tooling is good enough. ExplicitImports makes it not totally manual, but I still feel like it is not a pleasant workflow. I don't think we should drive package authors to a bad workflow. 
    - I think what might make it more doable is language server level integration so implicit -> explicit conversion is trivial and, more importantly, making it feasible to use names exported from packages that you haven't explicitly imported yet. During greenfield development, if every time you want to use a new name you need to first add the import by hand, that's a bad experience.
    - ExplicitImports has correctness issues because it works by re-implementing lowering in an ad-hoc way. That's kind of OK when you opt-into it, but I think it's an extra bad experience if you don't.
    - if we're going fully python style, we should support qualified names (`np.xyz`), but ExplicitImports doesn't have any functionality for converting to qualified names.
    - new developers especially will try to stick with the defaults and do what is recommended if they are trying to learn and be "good citizens" and they will also struggle the most with bad tooling
- The other side is that it's not really supposed to matter much where a name comes from, because it's behavior is defined by methods across packages. So asking "which package defines this" is sometimes the wrong question in Julia, and the right question is "what is it supposed to do", which is answered by its docstring. I think this is secondary though.

I think the "semver hole" of two packages exporting clashing symbols is a real problem, and I can't think of a solution besides qualified access or explicit imports, but I think nudging developers too hard towards explicit imports is not currently the right answer, because the tooling is immature so it'll be a bad experience.

fingolfin (changes requested):
Thanks for your contribution. However I am very sceptical about this change. IMHO having `ExplicitImports.test_explicit_imports` as a required check goes a step to far. @ericphanson puts it right: think the tooling is not good enough.

Using `ExplicitImports` is every unergonomic as the list of "TODO: fix instead of ignoring shows". The fact that you left that as a TODO instead of "just fixing it" is IMHO telling.

Also, since apparently `ExplicitImports.test_explicit_imports` exists, it doesn't seem to me we are adding much value anyway. If someone wants to use `ExplicitImports` but personally I wouldn't want to force this on everyone using Aqua at this point. I certainly would not want it for my packages.

fingolfin:
I find it sad that my points are mostly ignored... not refuted, but ignored:
1. the people coming here are self-selected: you find this checks useful, you are OK with being forced to spend hours and days adjusting your code to pass these tests, you already understand the fine points, you think "you just need to fix these things in Julia itself and then wait for a release there" is a useful way for you to spend time, etc. -- but maybe, just maybe, you can recall a time where all of these would have been unsurmountable barriers for you?
2. there are qualitative differences: ExplicitImports is a strategy for catching or preventing certain kinds of issues, but it requires rewriting code in ways that are arguably un-idiomatic these days (perhaps that means our "idiom" should change, fine, but can we at least agree this is qualitatively different from enforcing well-established standards like "type piracy is bad" ?
3. I love JET but fixing JET reported issues often is *hard*. I recently got half a dozen or so PRs into Julia just to get one particular package I am working JET-free, but for others we gave up because it would require changes to the Julia language on levels that are just way out of my reach (like... modifying the broadcasting code which is basically undocumented and undertested)
4. If people want this, I repeat my idea of making a HardcoreAqua / SuperAqua whatever umbrella package, and you can do what you like there. Seems lots of you are interested in that, so it shouldn't be a problem to get that started quickly, right?

Krastanov:
Maybe a relatively quick solution that will address Max's concerns would be to have a fairly centralized flag in Aqua that turns it "opinionated and harsh" and have it set to false by default but have it relatively loud in logs. E.g. if it is not set do `@info "You can enable much harsher and opinionated set of checks by doing so and so, but such checks are mostly directed to people maintaining foundational parts of the ecosystem or large organizations"`.

goerz:
If you want my two cents: Easy implicit imports were a huge mistake. I'd love if in Julia 2.0, `using Package` was a synonym for `import Package`. It took the Python ecosystem years to successfully discourage `from numpy import *` in packages, and it's a real shame that Julia didn't learn from that lesson.

It makes it impossibly hard to reason about any Julia code. Even for new package READMEs, usage example are often hard to understand, because I can't actually tell that a particular function is from the package if I don't have the public functions from Julia and any other package used in the example memorized.

So anything we can do to eradicate implicit imports from the ecosystem is a good thing. Please do make this a default.

PR #375, opened 2026-04-23
Symbol type parameters can be piracy

### LINKED ISSUE #374: `detect_piracies` misses some cases
The following is an example of piracy, but it is not caught by `test_piracies`:
```julia
julia> using TestEnv; TestEnv.activate(); cd("test")

julia> using Aqua

julia> module Pirate
       using IntervalSets

       Base.:+(iv::ClosedInterval, n::Real) = minimum(iv) + n .. maximum(iv) + n
       end
Main.Pirate

julia> Aqua.test_piracies(Pirate)
Test Passed

julia> using IntervalSets

julia> (1 .. 3) + 2
3 .. 5
```

timholy:
As described in #374, types like

    ClosedInterval{T} = Interval{:closed,:closed,T}

from IntervalSets blocked piracy-detection because `:closed` Symbol params caused `is_foreign` to return `false` unconditionally.

Fundamentally, this revealed problems with the heuristic used by `is_foreign`. This was essentially acknowledged as a problem in the comments.

The strategy taken here is to change the heuristic: if the combination `SomeType{:foo}` appears in the module that defines `SomeType`, then this combination is treated as a foreign type. If no such combination exists, then the Symbol is treated as a user-defined dispatch tag, and the type is treated as a user-defined type. To avoid the need for "grepping" through source code, there is one additional requirement for it to be treated as foreign: the type/symbol combination must be defined as a type alias (UnionAll) in the owning module.

Fixes #374.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Fix some world age issues in `test_ambiguities`. ([#366])
+- Symbols as type-parameters do not disable piracy checks in `detect_piracies`. ([#375])
 - The minimum supported julia version is increased to 1.6. ([#328])
 
 ## Version [v0.8.14] - 2025-08-04
--- a/src/piracies.jl
+++ b/src/piracies.jl
@@ -56,22 +56,22 @@ end
 is_foreign(@nospecialize(x), pkg::Base.PkgId; treat_as_own) =
     is_foreign(typeof(x), pkg; treat_as_own = treat_as_own)
 
-# Symbols as standalone type params (e.g. Val{:my_tag}) are treated as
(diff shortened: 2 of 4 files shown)


timholy:
I saw that the one failure was about a label. I don't have suitable privileges to add labels, but I did add an entry to the CHANGELOG.

lgoettgens:
Sorry for ignoring this for so long. I now had the time to finally review it. Unfortunately, this now has non-trivial conflicts due to https://github.com/JuliaTesting/Aqua.jl/pull/377. Could you try to resolve them?

One semantical question, because I am not sure what you are concretely trying to do here:
Let's assume that a dependency defines
```
ClosedInterval{T} = Interval{:closed,:closed,T}
```
and a function `foo`

As described earlier, defining `foo(::ClosedInterval)` is type piracy and `foo(::Interval{:myownsymbol,:bar})` is not. But what about `foo(::Interval{:closed,:closed,T}) where T`?

PR #377, opened 2026-05-18
Make `Aqua.test_piracy` recursive

### LINKED ISSUE #311: piracies detection only checks top module
I guess it should check also the inner modules, maybe with a "recursive" parameter

giordano:
Fix #311.  Implementation based on https://github.com/NumericalEarth/Breeze.jl/pull/713 and https://github.com/CliMA/Oceananigans.jl/pull/5602.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The minimum supported julia version is increased to 1.6. ([#328])
+- Make `test_piracy` check modules recursively for type piracy. This may result in failures for downstream users. ([#377])
 
 ## Version [v0.8.16] - 2026-06-05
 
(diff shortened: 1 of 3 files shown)


lgoettgens:
Thanks for your PR! Two comments:
1. it unfortunately has conflicts
2. Aqua already has a function `walkmodules` that could be used here, as it is quite similar. See https://github.com/JuliaTesting/Aqua.jl/blob/2f48bcff12950acb86f06a92e1f134a4458b1dd8/src/undocumented_names.jl#L25 for how it has been used in a different test that needs to recurse submodules

giordano:
I rebased on `master`, and used the existing `walkmodules`, but I removed the `recursive` argument I initially added to `test_piracies` because `walkmodules` makes a lot more annoying to support both a recursive a non-recursive version, and honestly I don't care too much about a non-recursive version (that's a footgun in my opinion)

giordano:
Bump

giordano:
Bump

lgoettgens (approved):
Thanks, looks good!

Note that this won't be backported to the current release branch, as this may result in more detected type piracy for downstream users, thus making their current CI setup fail.

PR #379, opened 2026-06-01
piracies: use Base.isType for Type checks

Keno:
Use Base.isType instead of reconstructing Type from parentmodule/nameof or Type internals when classifying method signatures. Prep for https://github.com/JuliaLang/julia/pull/61915.


Keno:
I need to get this merged and tagged to be able to PkgEval the Base change properly, so I'll merge this to master and have Claude do a separate backport to Aqua 0.8.*.

PR #380, opened 2026-06-02
[release-0.8] piracies: use Base.isType for Type checks (backport of #379)

Keno:
Backport of #379 to the 0.8 release series, plus the follow-up handling of non-DataType Type{Foo} in is_foreign_method.

Use Base.isType (with a fallback for Julia < 1.9) instead of reconstructing Type from parentmodule/nameof when classifying method signatures. This keeps piracy detection working when Type{Foo} is no longer a DataType on future Julia versions. Prep for https://github.com/JuliaLang/julia/pull/61915.

Also bump the version to 0.8.15.


Keno:
GitHub no longer has macOS 13 runners. Given that everything else is running and also passing on other OSes - I'll go ahead and merge this.

PR #389, opened 2026-07-21
Make persistent task test more robust

PatrickHaecker:
- Reports a precompilation failure as a precompilation error instead of misclassifying it as a persistent task.
- Set the default for `tmax` consistently to  30 seconds to reduce false positives when e.g. the system load is high.

Fixes #315, Closes https://github.com/JuliaTesting/Aqua.jl/pull/390.

PR #390, opened 2026-08-02
Report why the persistent-tasks probe failed instead of guessing

ChrisRackauckas-Claude:
## Problem

`test_persistent_tasks` spawns a wrapper package that `using`s the target package and, from inside precompilation, writes a `done.log` sentinel. If the subprocess exits without writing it, the check reports:

```
┌ Error: Unexpected error: /tmp/jl_XXXX/done.log was not created, but precompilation exited
Persistent tasks: Test Failed
```

and the test fails as though the package holds a persistent `Task`.

But the sentinel is also missing whenever the wrapper simply **failed to precompile**, which is a different problem with a different fix — and the reason is unrecoverable, because the child runs `Pkg.precompile(; io = devnull)`.

This matters most for the retryable `module is missing from the cache` race. That failure only *warns* and exits **0**, and the warning goes to `io`. With `io = devnull` the run looks like a clean exit with no sentinel, indistinguishable from a genuine persistent task.

I hit this downstream and it took a temporary patch to a CI job to recover the message. Once visible it was unambiguous:

```
34843.2 ms  ? jl_ji9KfebcbW
1 dependencies failed but may be precompilable after restarting julia
┌ jl_ji9KfebcbW
│  ┌ Warning: Module SciMLBaseDifferentiationInterfaceExt with build ID fafbfcfd-… is missing from the cache.
└  └ @ Base loading.jl:2643

done_log_written = false
exitcode = 0
termsignal = 0
```

Exit 0, no signal, no persistent task — a precompilation race. Before recovering that output I had ruled out a real persistent task, a native-library segfault, and memory exhaustion one at a time, because the check gives the same message for all of them.

## Change

Capture the child's precompilation output to a file inside the wrapper directory and, when the sentinel is missing, report it together with the subprocess exit code and termination signal, and note that a precompilation failure is the likelier cause:

```julia
@error "Unexpected error: … This usually means the wrapper package failed to precompile, \
        rather than that $pkgname holds a persistent task." \
       exitcode termsignal precompilation
```

Behavior is otherwise unchanged: the probe still returns `true`, output is still hidden on the happy path, and the "currently precompiling" branch is untouched. This only affects what is reported on failure.

Note the child's own stderr was already forwarded, so hard precompilation *errors* were partly visible; what was lost is everything Pkg writes to `io`, which is exactly where the missing-from-cache warning lands.

## Tests

New fixture `test/pkgs/PersistentTasks/FailsToPrecompile` that fails to precompile deterministically, plus a testset asserting the error names precompilation as the cause and carries `exitcode` and the captured output.

Full suite passes locally on Julia 1.12.6 — all 13 test files green, `test_persistent_tasks.jl` 12/12 (was 6).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The minimum supported julia version is increased to 1.6. ([#328])
+- `test_persistent_tasks` now reports why the probe failed when its wrapper package
+  never loads, including the captured precompilation output and the subprocess exit
+  status, instead of silently attributing the failure to a persistent task.
 
 ## Version [v0.8.16] - 2026-06-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The minimum supported julia version is increased to 1.6. ([#328])
 - `test_persistent_tasks` now reports why the probe failed when its wrapper package
   never loads, including the captured precompilation output and the subprocess exit
-  status, instead of silently attributing the failure to a persistent task.
+  status, instead of silently attributing the failure to a persistent task. ([#390])
 
 ## Version [v0.8.16] - 2026-06-05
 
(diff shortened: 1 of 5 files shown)


lgoettgens:
@ChrisRackauckas @ChrisRackauckas-Claude could you please explain the differences between this PR and https://github.com/JuliaTesting/Aqua.jl/pull/389? From the description, both seem to try to fix the same issue, but with different patches.

You may also add comments to that PR with things that should be changed there with an explanation.

ChrisRackauckas:
I didn't know #389 existed, and it's better

PR #392, opened 2026-08-30
Add test_workspace_compat guarding compat hygiene of test workspaces

cossio:
## Summary

This PR adds a new check, `Aqua.test_workspace_compat`, run by default as part of `test_all`.

Since Julia 1.12, a package can declare its test project as a workspace member:

```toml
[workspace]
projects = ["test"]
```

Workspace members share a single manifest, and Pkg **intersects** the compat bounds declared by every member when resolving it. A `[compat]` entry in `test/Project.toml` for a name the root `Project.toml` already owns therefore silently narrows the resolve: the root's declared bound stops being the effective one, and nothing warns about it.

This is not hypothetical: Dependabot introduces exactly this failure mode. When an update job is rooted at `/test`, it treats `test/Project.toml` as its own project file and synthesizes a `[compat]` entry for every direct dep that lacks one (the CompatHelper convention) — pinning deps the root already bounds, and shadowing the root's declarations from then on.

The new check tests that `test/Project.toml` declares no `[compat]` entry for any name the root project already owns: its `deps`, its `weakdeps`, anything in its `[compat]` (including `julia`), and the package's own name. Bounds for genuinely test-only dependencies (Aqua, SafeTestsets, ...) are legitimate and are not flagged.

### When the check applies

The check is a no-op unless **both** of the following hold:

- the root `Project.toml` declares `test` as a workspace member. Otherwise the test project resolves into its own manifest, where a repeated bound constrains only the test environment.
- the root's `julia` compat admits no version older than 1.12 (e.g. `julia = "1.12"`). Julia ≤ 1.11 ignores `[workspace]` and resolves `test/Project.toml` on its own, without inheriting the root's bounds, so a package that still supports those versions *needs* the repeated entries (see NumericalEarth/Breeze.jl#965, pointed out by @giordano). A missing `julia` compat entry counts as admitting old versions. In this case the check logs an `@info` message saying that it was skipped and why, so the no-op is visible in the test output rather than silent.

This double gating is why the check can be on by default: it only fires where the repeated bounds are genuinely redundant and shadowing. Today that is a small set of packages; it grows as packages drop Julia 1.11, which is exactly the point at which the repeated entries stop being load-bearing.

Example failure output:

```
PkgWithWorkspaceCompatConflict declares a compat entry in test/Project.toml for the following names already owned by the root project:
	Pkg = "1.5"	(the root project declares "1")
	julia = "1.6"	(the root project declares "1.12")
Workspace members share a single manifest, so these bounds are intersected and silently narrow the root's. Remove them from test/Project.toml.
```

The `@test` compares the offender list against `String[]` rather than asserting a bare `isempty`, so the failed test itself also names the offenders.

## API

- `Aqua.test_workspace_compat(pkg::Union{Module,PkgId}; broken = false, ignore = Symbol[])`, mirroring the conventions of `test_deps_compat`.
- `test_all(pkg; workspace_compat = ...)` keyword, `true` by default; accepts a NamedTuple of options like the other checks.

This upstreams [TestCompatHygiene.jl](https://github.com/cossio/TestCompatHygiene.jl) (same author), adapted to Aqua's conventions and gated as described above so it is safe to run by default. Happy to rename the check, make it opt-in, or change defaults if the maintainers prefer.

## Tests

- Unit tests on parsed project dicts for the helpers (`declares_test_workspace`, `admits_julia_before_1_12`, `root_owned_names`, `find_workspace_compat_conflicts`), covering pass/fail, `ignore`, sorting, the non-workspace no-op and the old-Julia no-op.
- Four new sample packages exercising the check end-to-end (conflict, clean workspace, compat outside a workspace, workspace but still supporting Julia 1.11), plus `broken = true` coverage and a check that the skip notice is logged.
- The whole test suite passes locally on Julia 1.12.7 and 1.10.12, and the docs build.

> Note for maintainers: the `changelog: added` label needs to be set by someone with triage permission (the changelog entry is included in the PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KYxRBkjYcsALngLXTshitE


cossio:
Could a maintainer please add the `changelog: added` label? I cannot set labels on this repository, and the "Check for blocking labels" CI job requires one. The changelog entry itself is included in the PR.

giordano (changes requested):
I don't think this is a good idea at all, at least until a package needs to support Julia v1.11-.  In Julia v1.11- compat bounds aren't inherited via `[sources]`, so that enforcing this can lead to actual errors.  Concrete case: https://github.com/NumericalEarth/Breeze.jl/pull/965#issuecomment-5543529084

cossio:
@giordano ok, would it help if we gate the test on supporting julia >= 1.12 ?

giordano:
Only if the package requires v1.12 explicitly so there's no way the package manager doesn't use workspaces, that may make more sense.

cossio:
Thanks @giordano, that is a real gap. I had only gated on the `[workspace]` declaration, which is not enough for a package that also supports Julia ≤ 1.11: there `[workspace]` is ignored and `test/Project.toml` resolves on its own without inheriting the root's bounds, so the repeated entries are load-bearing, as your Breeze.jl example shows.

[3b762fd](https://github.com/JuliaTesting/Aqua.jl/pull/392/commits/3b762fd3eee27a64c25713377895169ec0c26db2) gates the check additionally on the root's `julia` compat. It now only applies when that entry admits no version older than 1.12 (`julia = "1.12"`, `"1.13"`, `"1.12 - 1.14"`, ...; checked as `isempty(semver_spec(compat) ∩ semver_spec("0 - 1.11"))`, the same idiom `test_project_extras` uses for its 1.2 cutoff). A missing `julia` compat entry also counts as admitting old versions. For a package like Breeze (`julia = "1.11.9"`) the check is therefore a no-op, and it logs an `@info` message saying that it was skipped and why, so the skip is visible in the test output rather than silent.

That does make the check a no-op for most packages today; it starts to apply as packages drop Julia 1.11, which is exactly the point at which the repeated entries stop being load-bearing and become silent shadowing. If you would still rather have it opt-in (`workspace_compat = false` by default in `test_all`), I am happy to do that instead.

cossio:
@giordano yes, that's what I meant. This should be implemented now (also added tests that when supporting 1.11 this is a no-op)

PR #400, opened 2026-09-17
Skip uninstalled weak deps in `test_persistent_tasks`

### LINKED ISSUE #399: `test_persistent_tasks` errors on weak dependencies duplicated in `[deps]` (regression in v0.8.17)
Since v0.8.17 (#394), `test_persistent_tasks` errors for any package whose dependency tree contains a package that lists a weak dependency in **both** `[deps]` and `[weakdeps]` — the back-compat idiom for supporting Julia < 1.9. `SpecialFunctions` and `LogExpFunctions` both do this with `ChainRulesCore`.

## MWE

```julia
using Pkg
Pkg.activate(; temp=true)
Pkg.add(["SpecialFunctions", "Aqua"])
using Aqua, SpecialFunctions
Aqua.test_persistent_tasks(SpecialFunctions)
```

```
Test threw exception
  Expression: !(has_persistent_tasks(package; kwargs...))
  Unable to locate `ChainRulesCore`, a dependency of `SpecialFunctions`
  Stacktrace:
   [1] error(s::String)
     @ Base ./error.jl:44
   [2] visit(pkgdir::String)
     @ Aqua src/persistent_tasks.jl:102
   [3] manifest_entries(pkgdir::String)
     @ Aqua src/persistent_tasks.jl:107
```

Passes on v0.8.16, fails on v0.8.17 (reproduced on Julia 1.10 and 1.13).

## Cause
`manifest_entries`/`visit` iterates each package's `Project.toml` `[deps]` and hard-errors when `Base.locate_package` cannot find an entry. A weak dependency duplicated into `[deps]` is not installed, so it cannot be located.

Pkg keeps weak dependencies out of the manifest entry's deps, e.g.

```julia
[[deps.LogExpFunctions]]
deps = ["DocStringExtensions", "IrrationalConstants", "LinearAlgebra"]  # no ChainRulesCore
    [deps.LogExpFunctions.weakdeps]
    ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
whereas visit copies `prj["deps"]` verbatim and traverses all of it.

## Suggested fix
Mirror Pkg: subtract `weakdeps` from the entry's deps, and traverse weak
dependencies only when the environment actually installs them, so extensions
stay loadable and the check is not silently weakened.

```julia
for key in ("version", "weakdeps", "extensions")   # "deps" handled below
    haskey(prj, key) && (entry[key] = prj[key])
end
weakdeps = get(prj, "weakdeps", Dict{String,Any}())
deps = filter(((name, _),) -> !haskey(weakdeps, name),
              get(prj, "deps", Dict{String,Any}()))
isempty(deps) || (entry["deps"] = deps)
...
for (name, dep_uuid) in merge(deps, weakdeps)
    srcpath = Base.locate_package(PkgId(UUID(dep_uuid), name))
    if srcpath === nothing
        haskey(deps, name) || continue   # uninstalled weakdep ⇒ extension stays unloaded
        error("Unable to locate `$name`, a dependency of `$(prj["name"])`")
    end
    visit(dirname(dirname(srcpath)))
end
```
With this patch the MWE passes and `test/test_persistent_tasks.jl` stays green,
including the `dev`ed and path-tracked cases added in #394. Happy to open a PR.

I used the help of Claude to find the cause of the failing CI I have in https://github.com/JoshuaLampert/KernelInterpolation.jl/pull/207 and https://github.com/JoshuaLampert/SummationByPartsOperatorsExtra.jl/pull/108 and to create the issue. Others seem to have the same issue (see https://github.com/JuliaDEC/SimpleSplines.jl/pull/16).

lgoettgens:
Fixes #399.

Since v0.8.17 (#394), `manifest_entries` copies `[deps]` verbatim and errors on any entry it cannot locate. A weak dependency also listed in `[deps]` (the Julia < 1.9 compat idiom) is not installed, so the check failed for every package depending on e.g. `SpecialFunctions`.

Mirror Pkg: drop weak deps from the entry's `deps` and traverse them only when installed, so extensions stay loadable.

🤖 Created with the help of [Claude Code](https://claude.com/claude-code) (Fable 5.1)


giordano:
Can we get a new version very quickly pretty please?  #399 broke lots of packages (as you can tell from the back references to #399 and #394) 🥲

lgoettgens:
> Can we get a new version very quickly pretty please? #399 broke lots of packages (as you can tell from the back references to #399 and #394) 🥲

I am working on the backport release right now. Give me 5 more minutes, plus registration.

giordano:
I wanted to confirm that tests with Aqua v0.8.18 are [passing again for us](https://github.com/NumericalEarth/Breeze.jl/pull/1008), thanks for the quick resolution!

